PR #1, opened 2022-02-16
Update README to clone using https and not SSH

6by9:
This avoids having to have a Github account with SSH public key registered.


popcornmix:
+1

aallan:
Makes sense!

PR #3, opened 2022-02-17
Add a TensorFlow Example

aallan:
TensorFlow Lite inferferencing examples that carry out real-time object detection and labelling in a QT preview window using the [MobileNet V2](https://arxiv.org/abs/1801.04381) network trained on the [COCO dataset](https://cocodataset.org/#home). Still needs optimising but is good enough for an initial check-in!


examples/tensorflow/README.md
```diff
@@ -0,0 +1,66 @@
+# TensorFlow Lite Example
+
+TensorFlow Lite inferferencing examples that carry out real-time object detection and labelling in a QT preview window using the [MobileNet V2](https://arxiv.org/abs/1801.04381) network trained on the [COCO dataset](https://cocodataset.org/#home).
```
andrum99, line 3: `inferferencing` -> `inferencing`

examples/tensorflow/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Installing Picamera 2
```
andrum99, line 5: Remove the space between `Picamera` and `2`.

examples/tensorflow/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Ensure that you have installed the Picamera 2 library along with OpenCV.
```
andrum99, line 7: And again here.

examples/tensorflow/real_time.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2022 Raspberry Pi Ltd
+# Author: Alasdair Allan <alasdair@raspberrypi.com>
+# SPDX-License-Identifier: BSD-3-Clause
+
+# A TensorFlow Lite example for Picamera2 on Raspberry Pi OS Bullseye
+#
+# Install necessary dependences before starting,
```
andrum99, line 9: Extraneous comma at end of line

examples/tensorflow/real_time_with_labels.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2022 Raspberry Pi Ltd
+# Author: Alasdair Allan <alasdair@raspberrypi.com>
+# SPDX-License-Identifier: BSD-3-Clause
+
+# A TensorFlow Lite example for Picamera2 on Raspberry Pi OS Bullseye
+#
+# Install necessary dependences before starting,
```
andrum99, line 9: Another rogue comma

examples/tensorflow/real_time.py
aallan: it's not extraneous it's just a single run on sentence.

aallan:
Every so often "with labels" example crashes with an "index out of bounds" error caused by a race condition as the rectangles and texts arrays are updated by TensorFlow before this loop finishes. This doesn't happen for the non-indexed version without labels. Solution will be to rework the loop, append the label to the rectangles array rather than use a separate array. I'll fix the bug later today and repush.

examples/tensorflow/real_time_with_labels.py
aallan: Not rogue, intended.

PR #8, opened 2022-02-22
Initial video encoding support

chrisruk:


davidplowman:
This has now all been merged!

encoder.py
```diff
@@ -51,6 +51,8 @@ def format(self, value):
             self._format = V4L2_PIX_FMT_BGR24
         elif value == "YUV420":
             self._format = V4L2_PIX_FMT_YUV420
+        elif value == "XBGR8888":
+            self._format = V4L2_PIX_FMT_BGR32
```
kbingham, line 55: Is this correct? Below shows
   XRGB8888 == V4L2_PIX_FMT_XBGR32

So should this be 
elif value == "XBGR8888":
  self._format = V4L2_PIX_FMT_RGB32

or even 
   self._format = V4L2_PIX_FMT_XRGB32

I guess it depends on where the 'value' variable is coming from.

davidplowman: There is some confusion here. I believe the only 32-bit pixel format our V4L2 H264 encoder supports is this one (V4L2_PIX_FMT_BGR32), not any of the others (which I expect I'll have to add at some point). And when I configure libcamera (via the Python bindings) with 'XBGR8888' it works with this mapping and the reds and blues come out the right way round. Anything else either fails with an "OS Error" (i.e. not supported) or with the colours wrong.  So yes, there's something to look at but for the moment this works...

PR #9, opened 2022-02-22
Update drm_preview.py

tvoverbeek:
Add possibility to position the upper left hand corner of the preview window.


davidplowman:
Thanks!

PR #10, opened 2022-02-23
Make the call to `close_camera()` Idempotent

meawoppl:
If the user calls `.close_camera()` when the object goes out of scope the `__del__` method will call it again. and result in this error logging:
```
Exception ignored in: <function Picamera2.__del__ at 0x734164a8>
  File "/home/pi/picamera2/picamera2.py", line 51, in __del__
    self.close_camera()
  File "/home/pi/picamera2/picamera2.py", line 73, in close_camera
    self.camera.release()
AttributeError: 'NoneType' object has no attribute 'release'
```

I would strongly recommend against using the python __del__ machinery more generally, but that is a conversation for another day.

### DIFF
--- a/picamera2.py
+++ b/picamera2.py
@@ -68,10 +68,12 @@ class Picamera2:
         # Release this camera for use by others.
         if self.started:
             self.stop()
-        if self.verbose:
-            print("Close camera:", self.camera)
-        self.camera.release()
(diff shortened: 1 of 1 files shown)


meawoppl:
I move this to a branch from my work account, and reopened as #16

PR #12, opened 2022-02-23
Tidy Up Logging

meawoppl:
I can transition this to a more typical [python logger](https://docs.python.org/3/library/logging.html) if that is desired. In either case, this lowers code size and makes that transition easy later on if desired.


davidplowman:
Thanks for this, it is indeed tidier!

PR #16, opened 2022-02-24
Make Call to `.close_camera()` idempotent

meawoppl:


meawoppl:
Transferred from #10

meawoppl:
There is an option in github for "squash and merge" which will do that for you. I am also happy to do the rebase but it will have to wait until I finish some other bits.

PR #21, opened 2022-03-04
Context Manager and General Cleanup

IanTBlack:
-  Added self.camera_num  to class instantiation. This would mean that users would need to call and assign different names to a new Picamera2 class to use two cameras.
- Changed open_camera to open to stay in tune with start,, stop ,close etc. I don't know if it is bad form to use it in a class.
- Added __enter__ and __exit__ to allow for use of Picamera2 in a with statement.
- Created conversions.py for file or buffer conversions that don't necessarily require Picamera2.
- Changed camera_manager attribute to cm to reduce number of characters per line.
- Changed verbose_print to __verbose_print so it can't be called by a user.
- Changed start_, stop_ to __start, __stop so it can't be called by the user and to limit confusion when using interactive IDEs.
- Added initialize and acquire functions to give more flexibility to developers.
- Changed time.time() to time.monotonic() to remove interference with clock changes (although unlikely).
- Added reset_flags, which could probably be made private.
- Minor changes to messages displayed with errors.
- Added camera id (cid) and camera index (cidx) to initialize. This gives the class access to camera identifying information that the user may not know or may input incorrectly at instance creation.
- Moved self.sensor_resolution to still_configuration function because it had the only usage.
- Cleaned up some lines to bring them close to pep8.


IanTBlack:
I now see a mistake with moving self.sensor_resolution, so that can be moved back to open().

davidplowman:
Hi, thanks very much for doing all this. I'd like to adopt most of these changes, perhaps there are a few where we might think about exactly how to present them.

Mostly I think I'd like to split up the commits so that it's clear what each one is. Perhaps we could have a "PEP8" commit on it's own? Commits that just rename variables and functions might be separate from ones that rearrange code, perhaps. Also perhaps we could remove any diffs that do absolutely nothing (e.g. deleting newlines from the ends of files).

The purpose is simply to start making commits easier to review and to follow over time. I appreciate that things have been a bit random in the rush to get something out there for folks to try, but I would like to start improving the process a little. What do you think?

IanTBlack:
Can do.  I'll work on separating them out over the next couple of days. If you see anything else, let me know.

PR #23, opened 2022-03-07
Added context manager.

IanTBlack:
The __enter__ and __exit__ functions of a class allow operations to be wrapped in a with statement. Much like the original picamera, this would allow the camera to automatically open and close without the user explicitly doing so.

### DIFF
--- a/picamera2.py
+++ b/picamera2.py
@@ -48,6 +48,12 @@ class Picamera2:
         if self.verbose > 0:
             print(*args)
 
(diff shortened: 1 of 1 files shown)


davidplowman:
Thanks for this. In the previous set I think you had an example? That would be nice too!

IanTBlack:
Okay, I submitted another PR. 

What do you think about creating different branches for separating tasks or for delineating the purpose of the PR?

PR #24, opened 2022-03-07
Added context manager and context manager example.

IanTBlack:
I clicked revert from the previous pull request out of curiosity. This PR also includes an example on how to use the context manager built from the capture_headless example.

### DIFF
--- a/examples/context.py
+++ b/examples/context.py
@@ -1,0 +1,22 @@
+#!/usr/bin/python3
+
+"""
+In this example, a "with" statement is used to automatically open your camera
+and then close it after all the operations within the "with" statement
+are completed.
+"""
+
+from picamera2 import *
+from null_preview import *
+
+camera_num = 0  # Let's choose the camera at index 0.
+verbose = 1  # Let's also print informative messages to the console.
+
+with Picamera2(camera_num, verbose) as picam2:
+    config = picam2.still_configuration()
+    picam2.configure(config)
+    preview = NullPreview(picam2) # Don't preview the camera.
+    picam2.start()
+    np_array = picam2.capture_array()
+    print(np_array)
+    picam2.capture_file("context_demo.png")
--- a/picamera2.py
+++ b/picamera2.py
@@ -48,6 +48,12 @@ class Picamera2:
         if self.verbose > 0:
             print(*args)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self,exc_type, exc_val, exc_traceback):
+        self.close_camera()
+
     def __del__(self):
         """Free any resources that are held."""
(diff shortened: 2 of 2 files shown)


davidplowman:
Hi, I'm actually having a bit of trouble with this commit. I'm finding that the following:
```
from picamera2 import *
from null_preview import *
import time

with Picamera2() as picam2:
    config = picam2.preview_configuration()
    picam2.configure(config)
    preview = NullPreview(picam2) # Don't preview the camera.
    picam2.start()
    time.sleep(1)

with Picamera2() as picam2:
    config = picam2.preview_configuration()
    picam2.configure(config)
    preview = NullPreview(picam2) # Don't preview the camera.
    picam2.start()
    time.sleep(1)
```
causes a crash. To fix it you have to do a `preview.stop()` at the bottom of the first `with` block.

I think the problem is that even though the camera stops when we leave the `with` block, the event loop (which is run by the `NullPreview`) keeps going, and seems to create some `CompletedRequests` with a broken `Picamera2` in them. So presumably we'll need an automatic way to stop the event loop when we leave the `with` block.  The situation is different, of course, when Picamera2 runs as part of a Qt application and doesn't have a separate event loop. In that case, I guess the whole `with` idea probably isn't useful.

IanTBlack:
Ah. I see. I assumed it was an issue elsewhere because I had been been using Spyder .

I've tried to not use the NullPreview, thinking that it would still not do a preview, but it hangs up. 
Does NullPreview need to be given when a user doesn't want a preview? I believe this is significantly different from the original PiCamera, where the user did not need to explicitly decline a preview. Does the picamera2 use QT when no preview is given?

Now that I know that preview.stop() exists, I think the way to fix this would be to add flag a that tracks the preview status in the Picamera2 class. This would mean enabling something similar to start_preview and stop_preview like in picamera. The flag could then be wrapped in a try catch statement in __exit__ before the camera stop is givien. I think that would attempt to close the preview if it existed.

IanTBlack:
I'll work on this some more. I think I now understand the problem a little more. I am about 80% done with a solution that would make it easier for a user to establish a preview, but I've run into an OpenGL error that I am not familiar with.

The funny thing is that everything works fine in Spyder with no errors, but when I run it in terminal it hands up on a error associated with eglCreateWindowSurface.  My usual experience would be the opposite. 

I think it would better help me if I could reformat the acquisition and start procedures similar to the original PR. I think it would be useful to have individual functions for each action, but still maintain the higher level functions to make things easier for less experienced users in the future. If it's okay with you, I'll close this PR and will focus on that. I don't think the current placement of __enter__ and __exit__ will hurt anything.

IanTBlack:
I think that makes sense.  Sounds good to me.

PR #26, opened 2022-03-09
Add an encoder class that allows multi-threaded software encode

davidplowman:
We add:

MultiEncoder - a base class that can be derived from to implement
multi-threaded software encode.

JpegEncoder - an example using the MultiEncoder.

examples/capture_mjpeg.py - capture an MJPEG stream using the above.

### DIFF
--- a/examples/capture_mjpeg.py
+++ b/examples/capture_mjpeg.py
@@ -1,0 +1,21 @@
+#!/usr/bin/python3
+
+from null_preview import *
+from jpeg_encoder import *
+from picamera2 import *
+import time
+
+picam2 = Picamera2()
+video_config = picam2.video_configuration(main={"size": (1920, 1080)})
+picam2.configure(video_config)
+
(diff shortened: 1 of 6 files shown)


davidplowman:
@chrisruk Hi Chris, would you mind taking a look over this and see if it conforms to how you expect the Encoder class to be used, and is also in line with any work you're doing? Also any "prettifying" suggestions also welcome!

(Note that the hardware JPEG encoder is still important, and will give much better performance especially on older Pis, but some users might prefer a more "conventional" encoder.)

chrisruk:
Hi David, sure just having a look now, it does look in line with how i'm using the Encoder class.

I'd not used the threadpoolexecutor before, but using that along with the queue seems to preserve 
the job order for the output thread, which is neat :)

I'll try and create a hardware v4l2 jpeg encoder soon too, as iirc that was possible with a small change to the h264 encoder.

PR #27, opened 2022-03-09
Motion detection example

chrisruk:
Initial motion detection example (without using circular buffer currently)


davidplowman:
Oops, I think I marked this "ready for review" without realising. Anyway, it looks good to me so let me know when you'd like me to merge it!

chrisruk:
Cheers, yep, if you can merge that then, that'd be great

PR #28, opened 2022-03-10
Formatting, Logging, and Preview Changes

IanTBlack:
Sorry to keep dumping PRs on you. I closed the last two and wrapped their commits into the following. Hopefully I've made the commits granular enough. The have their own descriptions in addition to the book I have written below. If you require additional clarification or concerns, let me know.

**Repo Structure**
The biggest change is probably the locations of the files. My hope is that this will encourage a structure that gets picamera2 toward something that can be put on PyPI. I've added a folder for "tests". Right now that just holds a few scripts that you can use to test out functionality I added below. I've also moved everything to within its own picamera2 folder. So imports now must begin with picamera2.*next option*. Previews were moved to their own folder, as were the encoders. There were a couple of helper files that I wasn't sure where to put, so I put them in encoders. A utils folder now holds utils that may not necessarily pertain to the user. I also created a converters file next to the base picamera2 file because I think it might be something that users might want to use.

**Logging**
I added a basic logger that uses the native logging module to print messages to console. Message severity can be specified and depending on what the user chooses for verbose_console, only certain severities will be displayed on the console. The format is pipe delimited as datetime | message type | message. Datetime is ISO8601 compliant. The addition of the logging module also leave opportunity to create a log of the camera messages as a file or to incorporate them into the system log (which is functionality that logging says it can do, but I am not familiar). I can see this being useful for those who are looking to build scientific sensors around a RPI and camera (much like myself)>

**Function Names**
I changed some of the function names to be indicative of their "action_object". I also changed some names to have a '_' prefix so that they wouldn't immediately be available to a user.

**Previewing**
I created start_preview and stop_preview functions. I also added functionality to start_camera and stop_camera which will handle preview deletion and creation. A null preview is therefore created by default, so a user won't have to explicitly provide it. 
This should satisfy @davidplowman 's concern about the "with" block handling a preview. To start a preview, a user would then use the start_preview function and specify the type of preview as string. I was unable to get the DRM preview to work in the current release and this PR, so it is commented out. I would appreciate some insight there.

**Opening and Closing**
I modified the open and close functions to match their "action_object".  These functions are also built of lesser functions that a user can use if they want to get really specific with it. 

**__del__**
I commented out the __del__ function because I found it was interfering with subsequent runs of an example script. It would sometimes be called immediately after acquisition, so it would result an attempt to configure an unacquired camera.
Based on the [warning in the documentation](https://docs.python.org/3/reference/datamodel.html?highlight=__del#object.__del__), the __del__ function can be a bit hassle some. I found that when I would run an example in terminal, the completion of the module would clear everything anyway.

**Not all examples were updated.**
I didn't update all of the examples to match these changes. I'll wait to do that after your review.

**Tests**
I added some "tests" that showcase the functionality of the new additions. Something to note is that enabling and disabling QT preview is about 3-5 times faster than doing the same for QT GL.

Let me know if you have any concerns. If you would prefer that these changes be incremented over multiple PRs, let me know and I'll work on sorting it out.

- Ian


davidplowman:
Hi, thanks for doing all this, what a lot of work!

I think there's no reasonable way to review all this by eye-balling stuff, so I think what I'll do is copy the existing "main" branch into a branch called "next", and then merge all this onto that. We can then spend some time using it and changing bits as we see fit. Once all our ducks are in a row, all the examples and documentation are updated, I can push it onto "main".

I notice I seem to have created a couple of conflicts (sorry) - would you be able to patch those up and then I'll merge into "next".

Finally, given there's quite a lot of stuff here, would you be OK to confirm that you are donating your work to this project and to Raspberry Pi, which we shall continue to distribute under the license in the LICENSE file (BSD 2-clause licence).

davidplowman:
I think I've done all this now and pushed it up to "next". I'll start looking at it now. Some initial thoughts:

- I really like all the general reorganisation and work that's been done here.
- I was going to make a `start_preview` function myself, though I'm inclined to make it accept an actual preview object. That makes it easier for people to create new preview types, and saves us messing around with different arguments in the start_preview function.
- Making a preview when there isn't one may be wrong, for instance when running as part of a proper Qt app.

I'll begin by looking at the `start_preview` changes, and making all the existing examples work. Once I've done that I'll push it up here as a pull request for review. I'll close this PR now as it's been done (though onto the "next" branch for the time being).

Thanks! If you could still clarify on the copyright/licence issue I raised earlier, that would be appreciated.

IanTBlack:
@davidplowman Thanks for taking the time to take a look.

I have some concerns about passing a preview object through the start_preview function. I think that still requires a user to explicitly import a preview module in the custom script they may write. I think that may be conceptually challenging for younger or the least experienced users. Of course, I could be wrong. Succinct documentation and examples at release may be enough to relay the necessary steps.

I think the simplest thing to do would be to keep the initial argument as method. Then if it was supplied
as a string it would be assumed that the user wanted a particular (maybe Raspberry Pi supported?) preview.
But if someone wanted to pass their own custom preview, they could supply an object, which could be handled by start_preview with the following addition.

```
        elif isinstance(method,object):
            self._preview = method
```

But another option would be to create preview objects in scope of the picamera2 module (or the class?). These could be preloaded with default settings, which then the user can modify with width, height, x, y arguments. I do not know how reassigning width/height to a preview while the thread has started will impact performance, so that is something to consider.  

So just below the imports of the picamera2 module we could add...
```
NULL = NullPreview(*default settings*)
QTGL = QtGlPreview(*default settings*)
QT = QtPreview(*default settings*)
DRM = QtPreview(*default settings*)
```
Then the start_preview function could look like below.
The user would pass one of the objects above and the width, height etc if they didn't have a custom preview.
Then if the user were to pass their own preview object (while also setting the other arguments to None),
it would be assumed that it is a custom object and it would be set to _preview in its entirety.

```
    def start_preview(self,method =NULL,width = 600, height = 480,
                      x = None, y = None):
        """
        Width and height are used by all previews except for the null preview.
        x and y are only used by the DRM preview.
        """
        try:
            self._preview.stop()
            del self._preview
            time.sleep(1)
        except:
            pass
       if all(m is None for m in [width,height,x,y]):
               self._preview = method
               return True
       else:
               method.width = width
               method.height = height
               method.x = x
               method.y = y
               self._preview = method
               return True
```

Just a couple of thoughts I had about start_preview.


I think previously you had indicated that you were interested in making things easier for a less experienced user. I think this is a good goal and it resonates well with me when I think about it from an educational perspective. I think it is possible to achieve that while also providing functionality that suites the needs of advanced users and developers.

As for the license/copyright issue, I am fine with donating my work so long as picamera2 remains under BSD-2. I plan to put my own work, which will be built on top of picamera2, under MIT. Do you foresee any conflicts there? Thanks!

davidplowman:
Thanks for the feedback. I'm working my way through all those examples but mostly they're working without too much trouble. I'll post a PR once they're all good again and you can see what I changed.

I think we could do something like that with the start_preview function. Or perhaps have a "Preview" enum that you could pass in, or replace it with an actual object? So `picam2.start_preview(Preview.QtGl)` would be the same as `picam2.start_preview(QtGlPreview())`.

We'll continue to publish this under BSD 2-clause. It's very permissive (more so than GPL type licenses) as you can basically do whatever you want, people could even copy it, modify it, and make their version closed-source. The only constraint is that you have to reproduce (Raspberry Pi's) copyright notice.

Of course the nature of "donating" is that you don't strictly get to determine what happens afterwards... but I hope it's pretty clear that we have no interest in doing anything other than publishing as we are.

PR #29, opened 2022-03-14
Next

davidplowman:
This pull request builds on top of the existing changes on the "next" branch. Most of the changes, in particular the folder re-organisation, remain though some are modified (the `start_preview` function for example), and others are reverted for the moment pending further discussion (some of the function renaming).

Most of the commits on this branch leave the examples in a non-working state, and the documentation out of date, and fixing this up while keeping the commit history looks to be quite difficult, so all the changes on this branch will probably need squashing into a single commit before pushing onto main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,19 @@ sudo apt install -y libatlas-base-dev
 sudo pip3 install numpy --upgrade
 ```
 
+## Contibuting
+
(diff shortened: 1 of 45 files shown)


davidplowman:
I've actually pushed all the changes now, so closing this.

PR #42, opened 2022-03-16
Rename default_controls to camera_controls

davidplowman:
This is clearer and also matches camera_properties. (Note that
sensor_format and sensor_resolution refer specifically to the image
sensor and are therefore good as they are.)


davidplowman:
@chrisruk Hi Chris, another one for you - even more trivial this time!

PR #45, opened 2022-03-17
Qtgl fix

davidplowman:
These two commits improve the shutdown sequence of the QtGlPreview and DrmPreview windows, with the result that you can now successfully open a second one!


davidplowman:
@chrisruk Hi Chris, a couple more commits, at least these are fixing some genuine problems!

PR #47, opened 2022-03-20
Fix typo

davidplowman:
It's "self.log" not "self.logger".


davidplowman:
@chrisruk Another one!

PR #49, opened 2022-03-25
Overlays

davidplowman:
The commits here implement overlays, that is, the ability to draw alpha-blended overlays on top of the camera image in the various types of preview window, without changing the actual camera image underneath. This entire feature is made up of 6 commits:

1. The first commit implements overlays in the OpenGL preview window. There's also a very small amount of plumbing just to send the overlay from the Picamera object to the preview.
2. The second commit repeats this work for the simple Qt software-rendered version of the preview window.
3. Next we do it all again for the DRM/KMS version of the preview.
4. Finally we do the same for the `NullPreview`. Of course this doesn't do anything, we just want it to have the same API.
5. This commit adds some simple overlay examples and also puts them into the standard test list.
6. The final commit updates the docuementation in the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,30 @@ For no preview window at all, use `picam2.start_preview()` or `picam2.start_prev
 
 Please refer to the supplied examples for more information.
 
+### Overlays
+
+All the preview window implementations support simply overlays, which allows images with an alpha channel to be blended on top of the camera image. The facility is intended for adding simple graphics over the camera image rather than for complex high frame rate animations.
+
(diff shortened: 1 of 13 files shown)


davidplowman:
@chrisruk Here's a much less trivial pull request to look at when you have time! Probably a lot of it won't mean that much (TBH it wouldn't have meant much to me a week ago) but a second set of eyes is probably good nonetheless. Thanks!

chrisruk:
I've not used the OpenGL etc. bindings before, but been looking through the commits, which look good to me, will merge now. Cheers.

PR #50, opened 2022-03-25
Include EXIF data when saving JPEG files

davidplowman:
We add exposure time, ISO (gain), make and model.

Note that you will now need piexif (pip install piexif) in order to be
able to load picamera2.


davidplowman:
@chrisruk Hi Chris, another commit - a reasonably straightforward one though this time! Note that you'll need to do `pip install piexif` to make things work again once you've got this patch.

PR #52, opened 2022-03-28
Add tuning parameter to Picamera2 constructor

davidplowman:
This allows a custom camera tuning file to be named, or for a tuning
object (a custom Python dictionary) to be passed in.

An example (examples/tuning_file.py) has been added, and included in
the standard test list, and documentation added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,14 @@ picam2.configure(preview_configuration)
 picam2.start()
 ```
 
+## The Tuning File
+
+Being Python-based, _Picamera2_ is a good environment for inspecting and altering the _tuning files_ that Raspberry Pi ships for all its supported cameras.
+
+A _tuning file_ lists all the parameters needed for a specific camera to produce images of acceptable quality, configuring both hardware (specifically of the ISP, or _Image Signal Processor_) and also of the 3A and other real-time algorithms that run to control the behaviour of the camera system.
+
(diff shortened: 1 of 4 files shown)


tvoverbeek:
If you just want to use a different tuning file (i.e. for the NOIR version) you can also set the environment variable LIBCAMERA_RPI_TUNING_FILE which is picked up by libcamera.
Does using LIBCAMERA_RPI_TUNING_FILE interfere with the method in this PR?

davidplowman:
Well yes, it is the same underlying method underneath. So for the NOIR camera you could do:
```
LIBCAMERA_RPI_TUNING_FILE=/usr/share/libcamera/ipa/raspberrypi/imx219_noir.json python
```
or in the Python script
```
picam2 = Picamera2(tuning="/usr/share/libcamera/ipa/raspberrypi/imx219_noir.json")
```
or
```
tuning = Picamera2.load_tuning_file("imx219.json")
tuning['rpi.awb']['bayes'] = 0
```
Really I'm thinking ahead a little bit because I want to be a bit more flexible about tuning files and how we use them. In the future I want to implement the "raw re-processing" feature. This would let you collect raw files, manipulate them in some way (image stacking, for example) and then push them back through the ISP. But what camera tuning would you want to use? The default probably isn't what you want now, so it would be nice to have a means of tweaking the tuning on the fly. But there's still some way to go before we get there!

But note that the feature is implemented in such a way that if you don't use it, then it won't touch the value of LICAMERA_RPI_TUNING_FILE. So it shouldn't break any previous behaviour, it merely gives you new options.

PR #53, opened 2022-03-28
Initial setup.py

chrisruk:
There are a few things I'm not sure about -
* email address to use
* initial version number
* other requirements that need to be installed
* how we will 'require' the correct libcamera library 
* Need to also bundle test scripts

Thanks!

I just noticed the style checker failed, I am using the pre-commit hook suggested in the guidelines though

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,4 +23,4 @@ setup(name='picamera2',
       },
       packages=['picamera2', 'picamera2.encoders', 'picamera2.previews', 'picamera2.utils'],
       python_requires='>=3.7',
-      install_requires=['PyQt5', 'pyopengl', 'piexif'])
+      install_requires=['PyQt5', 'pyopengl', 'piexif', 'v4l2@git+https://github.com/RaspberryPiFoundation/python-v4l2.git#egg=v4l2'])
--- a/setup.py
+++ b/setup.py
@@ -23,4 +23,5 @@ setup(name='picamera2',
       },
       packages=['picamera2', 'picamera2.encoders', 'picamera2.previews', 'picamera2.utils'],
       python_requires='>=3.7',
-      install_requires=['PyQt5', 'pyopengl', 'piexif', 'v4l2@git+https://github.com/RaspberryPiFoundation/python-v4l2.git#egg=v4l2'])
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi Chris, thanks for this. Not quite sure what's up with the style checker, it's not actually complaining about anything, it's actually _failing_ (which I've never seen before). I'll try and pull your tree down today and see if I can reproduce the problem.

davidplowman:
I've been looking at the reason for the checkstyle failures. I think the problem might be that the script doesn't cope with merge commits. Could you perhaps try rebasing onto raspberrypi/main instead, and removing the merge commits? That might also make for a slightly more understandable commit history too. I'll have a look at making it produce a more comprehensible error message!

davidplowman:
Hi Chris, seems like "computer says yes" now, which is great. It's still marked as "work in progress" so let me know when you'd like to merge it. I'm happy to address those various other questions as and when we know the answers!

As regards the email address, should I set up a "picamera2" address here, and we can forward stuff from there to both you and me?

chrisruk:
Hi David, that sounds good about setting up an email address.

Will let you know when it's closer to needing to be merged.

chrisruk:
@davidplowman I've just been testing the setup.py script, first doing (listed here so I don't forget):

```
sudo apt install -y xserver-xorg xfce4 xfce4-goodies
sudo apt install python3-opencv python3-numpy opencv-data
```

In the picamera2 repo:

```
pip3 install . --user
export PYTHONPATH=/home/pi/Repositories/libcamera/build/src/py:/home/pi/Repositories/cam/kmsxx/build/py
./tools/run_tests.py -p ~/Repositories/picamera2/
```

All tests passed for me.  (I noticed once I got 'libEGL warning: DRI2: failed to authenticate' which happened when my xfce screen was 'locked')

davidplowman:
@chrisruk HI Chris, I'm happy to merge this - I'm guessing the history here isn't important so shall I squash it into a single commit?

chrisruk:
@davidplowman Cool, sure squashing it definitely makes sense. Thanks

PR #54, opened 2022-03-29
Circular buffer

chrisruk:
If you have time sometime, it would be good to have a chat about this @davidplowman 

I can't currently think of a way to synchronise the detection of motion on the 
picam2.capture_buffer("lores") data, with the data that the encoder is dealing 
with in the circular buffer.

Thanks!

### DIFF
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -32,8 +32,7 @@ while True:
         if mse > 7:
             if not encoding:
                 epoch = int(time.time())
-                encoder.output.dumpbuffer("{}-before.h264".format(epoch))
-                encoder.output.fileoutput = "{}.h264".format(epoch)
+                encoder.output.dumpbuffer("{}-before.h264".format(epoch), "{}.h264".format(epoch))
                 encoder.output.start()
                 encoding = True
                 print("New Motion", mse)
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -34,13 +34,13 @@ while True:
                 epoch = int(time.time())
                 encoder.output.dumpbuffer("{}-before.h264".format(epoch))
                 encoder.output.fileoutput = "{}.h264".format(epoch)
-                encoder.output.recording = True
+                encoder.output.start()
                 encoding = True
                 print("New Motion", mse)
             ltime = time.time()
         else:
             if encoding and time.time() - ltime > 5.0:
-                encoder.output.recording = False
+                encoder.output.stop()
                 encoding = False
     prev = cur
 
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -33,8 +33,7 @@ while True:
         if mse > 7:
             if not encoding:
                 epoch = int(time.time())
-                encoder.dumpbuffer("{}-before.h264".format(epoch))
-                encoder.output = open("{}.h264".format(epoch), 'wb')
+                encoder.dumpbuffer("{}-before.h264".format(epoch), "{}.h264".format(epoch))
                 encoding = True
                 print("New Motion", mse)
             ltime = time.time()
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -13,6 +13,7 @@ picam2.configure(video_config)
 picam2.start_preview()
 encoder = H264Encoder(1000000)
 encoder.buffersize = 30 * 5
+encoder.output = open("/dev/null", 'wb')
 picam2.encoder = encoder
 picam2.start()
 picam2.start_encoder()
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python3
+
+from picamera2.encoders.h264_encoder import *
+from picamera2.picamera2 import *
+import time
+import os
+
+lsize = (320, 240)
+picam2 = Picamera2()
+video_config = picam2.video_configuration(main={"size": (1280, 720), "format": "RGB888"},
+                                          lores={"size": lsize, "format": "YUV420"})
+picam2.configure(video_config)
+picam2.start_preview()
+encoder = H264Encoder(1000000)
+encoder.buffersize = 30 * 5
+picam2.encoder = encoder
+picam2.start()
+picam2.start_encoder()
+
+w, h = lsize
+prev = None
+encoding = False
+ltime = 0
+
+while True:
+    cur = picam2.capture_buffer("lores")
+    cur = cur[:w * h].reshape(h, w)
+    if prev is not None:
+        # Measure pixels differences between current and
+        # previous frame
+        mse = np.square(np.subtract(cur, prev)).mean()
+        if mse > 7:
+            if not encoding:
+                epoch = int(time.time())
+                encoder.dumpbuffer("{}-before.h264".format(epoch))
+                encoder.output = open("{}.h264".format(epoch), 'wb')
+                encoding = True
+                print("New Motion", mse)
+            ltime = time.time()
+        else:
+            if encoding and time.time() - ltime > 5.0:
+                encoder.output = open("/dev/null", 'wb')
+                encoding = False
+    prev = cur
+
+picam2.stop_encoder()
--- a/examples/capture_circular.py
+++ b/examples/capture_circular.py
@@ -12,7 +12,7 @@ video_config = picam2.video_configuration(main={"size": (1280, 720), "format": "
 picam2.configure(video_config)
 picam2.start_preview()
 encoder = H264Encoder(1000000, repeat=True)
-encoder.output = CircularFileOutput()
+encoder.output = CircularOutput()
 picam2.encoder = encoder
 picam2.start()
 picam2.start_encoder()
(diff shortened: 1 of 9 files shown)


davidplowman:
Hi Chris, thanks for these commits. I see the style checker is still failing. I'll try and have a look at that shortly, I just seem to be stuck in a rabbit hole to do with CMA allocations at the moment, but hopefully I'll emerge soon!

There are a few things I'd like to ensure with the circular buffer output. I think some of this may affect the way we implement encoder outputs, so we definitely need to put our heads together and think about it all!

Firstly, I think the whole circular buffer output needs to be completely independent of the codec. So for example someone could decide they want MJPEG in a circular buffer and it would "just work".

There are possibly some ramifications here: how does the output object know if it has a keyframe? Does the encoder have to tell it? Do we need an "output base class" that perhaps has an output(buffer, is_keyframe) method? So quite a lot to think about there. The V4L2 encoder tells us, I believe, if it made an I-frame, so we certainly need to make use of that.

What do you think?

I didn't quite understand the question about synchronising the lores image with what's being encoded. I thought the idea is that the circular buffer always contains "a couple of seconds of data" (or thereabouts) and when the lores image indicates motion we start saving everything, starting with those "couple of seconds". So I'm thinking that precise synchronisation (beyond a few frames) isn't really required. Have I understand that or is there something else going on? Thanks!

picamera2/encoders/h264_encoder.py
```diff
@@ -60,6 +61,16 @@ def _start(self):
         fmt.fmt.pix_mp.plane_fmt[0].sizeimage = 512 << 10
         fcntl.ioctl(self.vd, VIDIOC_S_FMT, fmt)
 
+        ext = v4l2_ext_controls()
+        extc = v4l2_ext_control()
+        extc.id = (0x00990000 | 0x900) + 226  # V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER
```
davidplowman, line 66: That long V4L2_CID thing probably wants to go into our v4l2.py, right?

Also, I think "repeat seq header" should be an optional parameter when the encoder object is created (along with the bitrate). We should probably also have the GOP size there too.

examples/capture_circular.py
```diff
@@ -0,0 +1,46 @@
...
+    cur = cur[:w * h].reshape(h, w)
+    if prev is not None:
+        # Measure pixels differences between current and
+        # previous frame
+        mse = np.square(np.subtract(cur, prev)).mean()
+        if mse > 7:
+            if not encoding:
+                epoch = int(time.time())
+                encoder.dumpbuffer("{}-before.h264".format(epoch))
+                encoder.output = open("{}.h264".format(epoch), 'wb')
```
davidplowman: I'm slightly feeling that we haven't separated "encoding" from "outputting" very cleanly here. I think this would be tidier if we hid all the "buffer dumping" and file handling in a custom output class. The functionality here seems fine, we just perhaps rearrange where some of it lives. I'd be tempted to:

1. Create a CircularOutput class. You create it with a file name, but it accumulates stuff in a circular memory buffer, and it has the "dump" function which writes everything to the file, For some applications, this class might be useful in its own right. It would also make for a reasonably simple demo of a "custom output".

2.  Then there's another custom output class, which incorporates the CircularOutput and file outputs together, but I can't think of a good name for it! When you create it, it makes a CircularOutput object which it holds and initially just passes it everything that arrives. It has a "start_saving" function which the application code here would call. This just sets a flag, and when the output function sees the flag is set, it dumps the CircularOutput and starts saving stuff to file for real. It could even re-open the same file for "appending" (or you could optimise that process by not actually closing the file in between).

What do you think?

chrisruk: I think this was a very good point, just been working on separating out the output functionality, which makes it nicer since it doesn't require using /dev/null, I've added a recording boolean that decides when frames are written.

chrisruk:
Oops, looks like I failed the tests :'(, will sort out next week

examples/capture_circular.py
```diff
@@ -0,0 +1,46 @@
...
+    cur = cur[:w * h].reshape(h, w)
+    if prev is not None:
+        # Measure pixels differences between current and
+        # previous frame
+        mse = np.square(np.subtract(cur, prev)).mean()
+        if mse > 7:
+            if not encoding:
+                epoch = int(time.time())
+                encoder.output.dumpbuffer("{}-before.h264".format(epoch), "{}.h264".format(epoch))
+                encoder.output.start()
```
davidplowman, line 36: Yes, I think this is much improved now that the work is being done by the output object. We should perhaps consider also adding an example that uses the JPEG encoder as this would prove that we've done the output vs. encoder separation correctly.

On a purely aesthetic note, the "dumpbuffer" function sounds to me like a bit too much of the implementation is showing through in the application code. Could we perhaps overload the start function and call only that (with any necessary extra arguments)?

I can see there is some appeal in dumping the circular buffer to file here in the application thread, as it doesn't then block the camera and encode threads. I have some slight misgivings about thread safety here, but I won't digress now, perhaps I'll put them in a comment at the bottom of the general discussion.

examples/capture_stream.py
```diff
@@ -18,12 +18,12 @@
...
-    picam2.start_encoder()
-    picam2.start()
 
     conn, addr = sock.accept()
     stream = conn.makefile("wb")
-    picam2.encoder.output = stream
+    picam2.encoder.output = FileOutput(stream)
+    picam2.start_encoder()
+    picam2.start()
     time.sleep(20)
```
davidplowman, line 27: Thanks for updating these examples. Actually I think the mjpeg_server.py example might need updating too - unfortunately it isn't part of the automated tests (because it runs forever), so you probably wouldn't have realised!

picamera2/encoders/h264_encoder.py
```diff
@@ -13,10 +13,12 @@
 
 class H264Encoder(Encoder):
 
-    def __init__(self, bitrate):
+    def __init__(self, bitrate, repeat=False):
         super().__init__()
```
davidplowman, line 17: We'll want a gop_size parameter too, but that's for a separate PR as it isn't related to this feature.

picamera2/encoders/h264_encoder.py
```diff
@@ -157,9 +170,9 @@ def thread_poll(self, buf_available):
                         # Write output to file
                         b = self.bufs[buf.index][0].read(buf.m.planes[0].bytesused)
```
davidplowman, line 171: At some point we should probably experiment with mmaping the DMABUFs that we get from the encoder. The function that gets this buffer (the outputframe method) would be responsible for copying it if they need to. But again, that's for a subsequent PR, not this one!

picamera2/encoders/h264_encoder.py
```diff
@@ -157,9 +170,9 @@ def thread_poll(self, buf_available):
...
                         self.bufs[buf.index][0].seek(0)
-                        if self._output is not None:
-                            self._output.write(b)
-                            self._output.flush()
+                        naltype = b[4] & 0x1F
+                        keyframe = (naltype == 0x7 or naltype == 0x8)
```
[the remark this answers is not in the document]
chrisruk: Good point, I've changed it now to get the flag from v4l2.

examples/capture_stream.py
chrisruk: Yup, missed that one, tested that after a small change to it

picamera2/encoders/h264_encoder.py
chrisruk: Good point, just created an issue for that so I don't forget

picamera2/encoders/h264_encoder.py
chrisruk: Yep, that makes sense, mentioned that in the h264 encoder issue I created, so as not to forget

examples/mjpeg_server.py
```diff
@@ -85,7 +85,7 @@ class StreamingServer(socketserver.ThreadingMixIn, server.HTTPServer):
 picam2.start_preview()
 picam2.configure(picam2.video_configuration(main={"size": (640, 480)}))
 output = StreamingOutput()
-picam2.start_recording(JpegEncoder(), output)
+picam2.start_recording(JpegEncoder(), FileOutput(output))
```
davidplowman, line 88: Nothing that needs changing, but a question of style for our users. Do you think they'd do better to create their own regular output class (`StreamingOutput` here) and wrap that in a `FileOutput`, or should we advise them to derive `StreamingOutput` directly from our base output class?

But let's leave this for now. If we think the alternative style is nicer we can change it later.

picamera2/encoders/output.py
```diff
@@ -0,0 +1,111 @@
...
+                self._fileoutput.write(frame)
+                self._fileoutput.flush()
+
+    def stop(self):
+        """Close file handle and prevent recording"""
+        self.recording = False
+        if self._circular is not None:
+            for frame, keyframe in self._circular:
+                self._fileoutput.write(frame)
+                self._fileoutput.flush()
```
davidplowman, line 110: The flush seems a bit unnecessary as we're about to close the file anyway, and we might actually slow it up by forcing smaller chunks to disk than is optimal. But I'd be surprised if there was a big effect so I can't say I'm too bothered...!

davidplowman:
@chrisruk I think this is looking good to me now. Shall I go ahead and merge it? Just the usual question... do you want the commit history here appearing or would you like to tidy them up at all? If you prefer I'll happily squash and merge them into one. Just let me know!

PR #58, opened 2022-04-04
Changes to rgb stream alignment and addition of rgb2* functions.

IanTBlack:
Taking a break from the mechanical design of my camera, so I should have more time to contribute in the future.

I was getting reduced size images when creating full size images on the IMX477. I changed align_stream to estimate the stream size to be high so that no pixels are lost.

The .npy is one of many ways to capture RGB values in a non-visual format. Will probably not be used by the majority of users, but is nice to have.

It might be worth it to move the converters back into picamera2.py and within the main class. Additions of metadata as exif and tiff tags could utilize some of the private information in the main class. This may require some reworking of some of the capture functions. The alternative is to require the user to explicitly pass metadata.

Signed-off-by: Ian Black <blackia@oregonstate.edu>

### DIFF
--- a/examples/app_capture.py
+++ b/examples/app_capture.py
@@ -3,15 +3,15 @@
 from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtWidgets import *
 
-from picamera2.previews.q_gl_picamera2 import *
-from picamera2.picamera2 import *
+from PiCamera2.previews.q_gl_PiCamera2 import *
+from PiCamera2.PiCamera2 import *
 
 
 def request_callback(request):
     label.setText(''.join("{}: {}\n".format(k, v) for k, v in request.get_metadata().items()))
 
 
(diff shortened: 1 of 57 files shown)


davidplowman:
Hi, thanks for these contributions. I'm trying to be more careful now about testing and reviewing changes now, I think things were a little bit haphazard at the very start. Anyway, I put some guidance on the README page ([here](https://github.com/raspberrypi/picamera2#contributing)). With that in mind, might you be able to:

1. Fix up the style checker complaints. I'm trying stick with PEP-8 wherever possible, though there will be occasions when different formatting is tidier in which case I'm happy to ignore it.
2. Look at any failing tests.
3. I'm trying to stick with "one change per pull request", so I'm wondering if changes to image sizes and new conversions should be separate PRs.
4. Also could you remove any unwanted commits, just so that what is left is as clear as possible?

Thanks!

PR #61, opened 2022-04-06
Add offset and window frame controls to preview windows

davidplowman:
The Qt-based preview windows now have x and y parameters so that users
can force a particular offset. Also the "frameless" parameter allows
the window to appear without a surrounding frame. The previous
behaviour (with a frame) is preserved as the default.

An extra example of this and some extra documentation are also
provided.


davidplowman:
I've removed the control to create "frameless" windows. I'm not sure it's useful, and it's easy to add if it turned out anyone wanted it. I've left the x/y window offsets as I think we probably should provide those (indeed the DrmPreview has them already).

Issue #62, opened 2022-04-10
Make q_picamera2 and q_gl_picamera2 dynamic size


jlprojects:
I think the camera preview in the q_picamera2 and q_gl_picamera2 should be made (optionally) to automatically fill the size of its container. This problem can be seen in examples/app_capture.py - resizing or maximising the window doesn't look very good. The preview in many application cases I'd expect will need to fit the maximum space available.

My project is a cine film scanner using the HQ camera. It's original control UI was written using picamera and tkinter which sort of worked, but very clunky. I decided to start again using picamera2 and PyQt5, so I'm fumbling to implement a new GUI.
The q_picamera2 and q_gl_picamera2 widget looks ideal for showing a live view for focussing and aligning the camera, and indeed I'm starting by building on app_capture.py for a live preview and basic camera settings.

I altered  q_picamera2.py to implement a dynamic resize which seems to work - patch below. It implements a resize event, decouples the camera image size from the label size, centres the camera image in the label. A layout control fills the label to the size of the parent widget. The overlay is handled and resized as necessary. It can be tested in examples/app_capture.py by replacing the`q_glpicamera2.py` import  to `q_picamera2` and the `qpicamera2` declaration to `qpicamera2 = QPicamera2(picam2,resizeable=True)`

I'm sure it can be optimised. Similar functionality should exist on the GL preview widget but I got lost trying to understand the GL code.  Just getting up to speed with basic PyQt at the moment. 

```
--- q_picamera2.py.orig	2022-04-10 09:53:31.931882445 +0100
+++ q_picamera2.py	2022-04-10 11:10:19.290102418 +0100
@@ -1,17 +1,29 @@
 from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtCore import pyqtSlot, QSocketNotifier
-from PyQt5.QtWidgets import QWidget, QApplication, QLabel
+from PyQt5.QtWidgets import QWidget, QApplication, QLabel, QVBoxLayout
 from PIL import Image
 from PIL.ImageQt import ImageQt
 import numpy as np
 
 
 class QPicamera2(QWidget):
-    def __init__(self, picam2, parent=None, width=640, height=480):
+    def __init__(self, picam2, parent=None, width=640, height=480, resizeable=False):
         super().__init__(parent=parent)
         self.picamera2 = picam2
+        self.image_ratio = height / width
+        self.img_size = QtCore.QSize(width, height)
         self.label = QLabel(self)
[...]

PR #65, opened 2022-04-12
Attempt at making dynamic sized qpicamera2 widget

jlprojects:
This is a first attempt at making the PyQT5 picamera2 widgets dynamically resizeable. In many applications a camera preview should fill the maximum available space in an application layout.
At the moment I've only changed the non-gl QPicamera2 widget - but ideally the GL and non GL versions should behave in the same way.
https://github.com/raspberrypi/picamera2/issues/62


davidplowman:
Hi, thanks for sending this! A had just a couple of initial questions:

1. In order to make this resizable, I see you've added an extra QVBoxLayout. Can it be done without this, or is this necessary - I'm new to PyQt so this is a genuine question!
2. Assuming we need the QVBoxLayout, we now have two slightly different implementations here, one with the QVBoxLayout, one without. Is there any advantage in the "without" version, or could we do away with it so as to have a single implementation?

jlprojects:
Hi,
1. This is my first foray into PyQt as well. I can't say I'm that proficient in Python, either. The QVBoxLayout is used so the preview image is centred automatically in the placeholder widget. I expect there's a way of optimising this implementation quite a lot. :)
2. I'm aware of the two implementations at the moment - I've not touched the GL version. That scared me somewhat, so thought I'd just try it out on the true PyQt version as a proof of concept. It would make sense to me to have one implementation that can automatically select the display method to use. 

Noted the style-check errors - I'll attend to those presently.

I'm currently at the learning curve stage of recreating my telecine application using Picamera2 and PyQt - learning along the way. Mainly painfully keep trying stuff until it starts working. Using qtcreator to lay out the UI without hand-coding it has been a boon - here's a screenshot showing the qpicamera2 widget being used as a preview:
resizeable-qpicamera2
.

jlprojects:
While reimplementing my app and learning a little more about PyQt5, it occurred to me it that QPicamera2 might be better implemented as a QGraphicsView object. The preview image stored as a pixmap in the scene. Qt then handles all the view scaling automatically. Will experiment and report back...

jlprojects:
Had a play around with the non GL QPicamera2 widget. I made it into a QGraphicsView - the preview is rendered into a QGraphicsScene - resizing is handled by Qt. Which seems to work nicely.

The overlay is added to the GraphicsScene when it is enabled, so it's not necessary to repaint it every time a new frame is available from the camera - that's handed by Qt. 

I also added an example based on app_capture.py with an additional checkbox to show an overlay. Performance seems fine on my Pi4.

davidplowman:
Thanks for sending this. I've got a bit of post-Easter catching up to do but will get onto it after that!

jlprojects:
Thanks - I'm sure it'll need some work. Found an issue when I integrated it into my app - that switching away to stills a configuration would send an oversized image to the preview, resizing the GraphicsScene. So I reinstated the height and width parameters to the constructor, and put in a check for the image size from the request. The size in the constructor should match the preview size, otherwise each frame gets resized. I think this is what is happening in the original version in the make_image call.

jlprojects:
I've had a go at bringing up this patch to the current main branch of Picamera2 - made a hash of doing so, but I think it's OK. 
Qt's fitInView method for resizing the graphics scene to the qgraphicsview has a fixed 2 pixel margin baked in. So I've had a go at making a reimplementation. Moving it into the resize event makes things more responsive - it's only necessary to rescale the scene when the window changes size.

davidplowman:
Hi, thanks for updating this pull request. I'm sorry that there's been quite a lot of churn over the last week, but I hope that may calm down a bit...

One thing to notice is that the old Qt preview code is gone, replaced by a unified version ([here](https://github.com/raspberrypi/picamera2/blob/main/picamera2/previews/qt_previews.py)), so you probably want to factor that into your change too. (I think I'd mumbled something a while back about wanting to do that!)

I'm also hoping I might at some point have some time to look at the GL version. Maybe we get this version to a state where it does what we want, then the trick will be to make the GL one behave the same. Also really appreciate that you've added a new example... thanks!

jlprojects:
Hi, thanks for the feedback.
The churn's to be expected :)  Also had an SD card meltdown on me at the weekend and had to recreate a new one from scratch. At least the picamera2 library's dependencies are much easier to install than before. :) Also as an FYI datapoint, it's on a 4GB Pi 4 with 64 bit Raspberry Pi OS, with updated Numpy and OpenCV installed via pip3. The picamera2 code seems to be working very reliably - the image processing and capture is all done with 12bit RAW data demosaiced with openCV. 

This version of the q_picamera2 widget seems to be working well with the film scanning application at least - it's inserted into a UI made using qtcreator.  The background colour of the QGraphicsView can easily be changed. A neutral grey is good when film scanning. Performance seems pretty good even over SSH.

I'll next have a look at seeing how it integrates into the unified Qt preview code. It looks like that's using the q_(gl)picamera2 widgets to display the preview?

As regards the GL version - I'm unsure how that works at present. So as you suggest I'm happy to work on the non-GL version for the moment and get that into shape
screenshot-2022-05-09
.

davidplowman:
Hi, I've tried to pick up this thread of work again. I've rebased your latest version on top of the latest Picamera code, you can find it [here](https://github.com/davidplowman/picamera2/tree/resizeable-qpicamera2). Hopefully I got it right, it wasn't short of conflicts! I also notice that the pesky border still seems to be there, I can't believe how hard it is to kill it!!

I've also had a go at some equivalent stuff for the QGlPicamera2 widget, you can find that [here](https://github.com/davidplowman/picamera2/tree/qtgl-improvements). In many respects it seemed easier because a lot of the OpenGL stuff just goes right underneath Qt's radar, so it can't get in the way. I also added an option to ignore the aspect ratio (just fill the window), and for a background colour when things get letter/pillar-boxed as it was (a) easy and (b) seemed like it might actually be useful.

I should also take a look at the DRM/KMS preview. There are no resize events there, of course, but it could still handle aspect ratios and overlays better.

jlprojects:
Hi - My pull request code is now up-to-date with main and seems to be functioning. Will have a look at the openGL improvements, and implement the options to ignore the aspect ratio and change the background colour to match. The latter can be done simply as part of the base QGraphicsView widget, but no reason it can't be done explicitly. 
I use this in my application - the commented line was used while debugging:
```
# self.qpicamera2.setStyleSheet("border: 4px solid red;")
self.qpicamera2.setBackgroundBrush(QBrush(Qt.darkGray, Qt.SolidPattern))
```
the border on the QGraphicsView is hard-coded 2pixels in the QGraphicsView fitInView method, so the workaround is to override it.  This is only called on a resize event, so not very expensive.

davidplowman:
Hi again, I've been doing some more work on all this and have created (yet another... sorry!) branch [here](https://github.com/davidplowman/picamera2/tree/preview-rework2).

On here I have all my work on the GLES and DRM/KMS versions, and I've overwritten the regular Qt version with yours, having done some more tweaking on it. It now avoids resizing the overlay to anything intermediate, and draws and clears overlays even when the camera isn't running. It also copes with stopping the camera, changing the image aspect ratio and restarting it. So it seems to be behaving pretty well to me.

Perhaps you could let me know if it seems to be behaving for you, I think it might be quite close to something one might be able to commit. I'd also like to attribute the commit where the QPicamera2 gets converted to a QGraphicsView to you as you've done most of the running on that, but we can talk about that in due course.

Thanks again!

davidplowman:
I think we've done all this now so I'll close it. If anything else comes up, please open another issue - and thanks for all your help with the one!

PR #72, opened 2022-04-22
Make the Qt signal/slot mechanism easier to use

davidplowman:
Adding a "signal_done" function and corresponding signal to the Qt
widget classes makes it much more convenient to request a Qt callback
when a camera operation completes.

There's a new version of app_capture.py, app_capture2.py, that uses
this mechanism to re-enable the push button once the capture operation
is done.


davidplowman:
Yep, I'll fix the conflict. Actually I seem to have got some logic slightly wrong here. I'll re-post it tomorrow. Thanks!

PR #74, opened 2022-04-25
Make Picamera2.start() create the event loop automatically

davidplowman:
This saves user's scripts from having to remember, but it does mean
that if you want a preview window, you should create it before
calling Picamera2.start().


davidplowman:
Still not sure about this. It replaces the awkwardness where you _have_ to call `start_preview` with the awkwardness that you sometimes _don't_ have to call `start_preview`...

meawoppl:
I for one think this is great. That said, my application is headless :stuck_out_tongue:

davidplowman:
Well, I think there seems to be a general consensus here slightly in favour of this change. @chrisruk Do you have any opinion either way?

chrisruk:
Yup, I'm happy with not having to manually call start_preview.

PR #76, opened 2022-04-26
WIP: Support for suggesting particular sensor modes

davidplowman:
Not sure about this functionality. I'd like it to be easier for applications to use and manipulate sensor modes, but is this helpful? It would enable things like:
```
config = picam2.preview_configuration(raw=SensorMode.FULL)
picam2.configure(config)
```
or
```
config = picam2.still_configuration(raw={})
update_sensor_mode_format(config['raw'], {'packed': False})
picam2.configure(config)
```
@naushir Do you think there are any other useful features we could add? Do you think we could do any of this better?

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -10,6 +10,7 @@ import time
 import tempfile
 import json
 from picamera2.utils.picamera2_logger import *
+from picamera2.sensors import *
 from picamera2.previews.null_preview import *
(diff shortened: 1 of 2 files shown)


naushir:
Can't think of anything that comes to mind right now.

It's not great that we have to keep a per-sensor mode list updated.  Is there no way we can do what libcamera-apps does to derive this list?

PR #82, opened 2022-04-30
Import cleanup

meawoppl:
Cleanup imports a bit. This includes killing `*` imports, grouping them as 

1. Builtin
2. External
3. This module

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1,23 +1,24 @@
 #!/usr/bin/python3
 
+import json
 import os
-import libcamera
-import numpy as np
-import threading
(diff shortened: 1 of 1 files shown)


davidplowman:
Thanks for this. Would you be OK to sign your name in the commit message according to the [guidlines](https://github.com/raspberrypi/picamera2#contributing)? Thanks!

meawoppl:
@davidplowman is that the commit format you are looking for?

davidplowman:
Yes, that looks fine. The 2 test failures are things I'm looking into, they're a problem down in libcamera so we can ignore them here. Thanks!

PR #83, opened 2022-04-30
Async Cleanup

meawoppl:
Setting attributed on parent objects is kinda confusing, so I think this makes the control flow a little cleaner and more robust generally.


davidplowman:
Hi, this seems like a good idea to me. I think it could go into just a single commit - could you maybe squash them together with a single comment and re-push? It might also pass the style checker that way (not sure why it got upset). Thanks!

PR #84, opened 2022-04-30
Add some typing

meawoppl:
Add a little bit of typing to the stack.


davidplowman:
I'm guessing all the failures here might just be because I need to "pip install" something on the automatic test machine? If you let me know what, then I'll do that and re-pushing will cause them to run again. Also, I'm thinking all these commits could be squashed together, and perhaps you could fix the little style niggle too? Thanks!

davidplowman:
Actually, if we do need to install something extra, could you put that in the installation instructions in the README file too? Thanks.

meawoppl:
Interesting. What version of Python are these getting tested against?

davidplowman:
I'm just going to close and re-open this so that the tests run again.

davidplowman:
So the test machine is reporting this:
```
  File "/home/pi/_work/picamera2/picamera2/examples/overlay_null.py", line 3, in <module>
    from picamera2.picamera2 import *
  File "/home/pi/_work/picamera2/picamera2/picamera2/picamera2.py", line 39, in <module>
    class Picamera2:
  File "/home/pi/_work/picamera2/picamera2/picamera2/picamera2.py", line 560, in Picamera2
    def get_completed_requests(self) -> List[CompletedRequest]:
NameError: name 'CompletedRequest' is not defined
```

meawoppl:
Ahh, I see the issue. 

As an aside, I am curious why you decided to build your own test/lint infrastructure. Is there a way I can run these tests locally or ideally see the stack trace from CI failures?

I would also be happy to make a Pytest drop in runner for your dedicated CI node. LMK

davidplowman:
You can run the tests locally just by running `tools/run_tests`. No reason for writing my own little test script except that it was quick and I don't know what actual Python programmers do.

I'd be open to using something better, probably best to discuss first and understand what improvements will come from it.

meawoppl:
I would say that most people using python use either [pytest](https://docs.pytest.org/en/7.1.x/) or [nose](https://nose.readthedocs.io/en/latest/).

These frameworks are really nice for a few reasons, the biggest is that they capture all sorts of state from the test (stdout, logs, timing, etc) and display it nicely when there is a failure. I would reccomend `pytest` largely because it does really nice things to unroll assertions like this:
image

I think all you would have to do to get it running against the current stack is:
- Install pytest
- Rename the `main()` functions in the tests folder to `test_something_something()` so pytest will discover/run them
- Call `pytest tests/*` (maybe with a few more flags)

meawoppl:
On a related note, basically everyone has standardized on the [**black** python formatting tool](https://github.com/psf/black). It's zero-config and -fixes- style changes for you, which is the major draw.

meawoppl:
I'm happy to make a quick PR of adding either of those to the stack here, just lmk your opinions/feelings.

meawoppl:
Squashed and signed as desired.

davidplowman:
Thanks for this latest version, it's looking pretty good now. I notice there are some test failures - the one that says "unicam has timed out" is nothing to worry about, we're working on that down in libcamera itself. Are the other ones working for you?

davidplowman:
Closing in order to re-open and re-run the tests (I've fixed the unicam timeout issue).

davidplowman:
I pulled down your branch and `tools/run_tests.py` does indeed report 3 failures. If you try the failing tests (for example, just execute `examples/opencv_face_detect_2.py`), they're missing the `time` import as a result of the file re-organisation.

Might be nice to mention to splitting the file up in the commit message, or even to have 2 commits (one for the file split, the second to add the typing), but I'm not too bothered on that. If you can fix the test failures I'm happy to merge.

meawoppl:
Yuck. This is why * imports are [generally not used](https://stackoverflow.com/questions/2386714/why-is-import-bad).

Let me do some modernization here and see if that helps. The tests seem to assume a specific layout of folders that I don't have but that is a PR for another day :/

meawoppl:
Oui, yeah. There are a bunch of places where things are getting stealthily imported through the `from picamera import *`.
For example, here:
https://github.com/raspberrypi/picamera2/blob/main/examples/opencv_mertens_merge.py#L6

Is actually responsible for `np` being defined :frowning:

meawoppl:
This one is gtg. I swept up the * import related issues in #95

PR #85, opened 2022-05-01
Use enums for camera controls

davidplowman:
Enums have been added for camera controls so we don't have to use
"magic numbers" any more.


IanTBlack:
Hi David,

I am working on some changes that I think would make it easier to navigate the controls. I don't think they should impact your addition of enums too much, but just wanted to make you aware.. I should have the PR ready by the end of the day today or tomorrow morning Pacific Time.

davidplowman:
One of the tests here seems to be failing, but I don't think it's "real". At least, not in the sense of being related to the change in this PR. I think a libcamera update is causing it, possibly because we're "sleeping" with the event loop not running which makes it complain. Will have to put that on the list of things to investigate, but it shouldn't hold up merging this PR.

PR #86, opened 2022-05-02
Added controls functionality.

IanTBlack:
This adds the controls module and Controls class.

This class handles controls in Picamera2(), providing several methods for issuing control changes to the camera.
Please see[ Jupyter Notebook](https://github.com/IanTBlack/picamera2/blob/controls/examples/notebooks/controls_example.ipynb).

There are also some commented lines in the controls module for documentation. Should those be taken out in favor of a ReadtheDocs?

It would also help to confirm the datatypes that libcamera requires for each control.
Reverted imports change to reduce conflict with PR #82 

Thanks!
Signed-off-by: Ian Black <blackia@oregonstate.edu>

### DIFF
--- a/examples/capture_headless.py
+++ b/examples/capture_headless.py
@@ -1,4 +1,4 @@
-#!//usr/bin/python3
+#!/usr/bin/python3
 
 from picamera2.picamera2 import *
 
--- a/examples/exposure_fixed.py
+++ b/examples/exposure_fixed.py
@@ -11,5 +11,6 @@ picam2.start_preview(Preview.QTGL)
 preview_config = picam2.preview_configuration()
 picam2.configure(preview_config)
 
-picam2.start({"ExposureTime": 10000, "AnalogueGain": 1.0})
+picam2.set_controls({"ExposureTime": 10000, "AnalogueGain": 1.0})
+picam2.start()
 time.sleep(5)
--- a/examples/notebooks/controls_example.ipynb
+++ b/examples/notebooks/controls_example.ipynb
@@ -16,7 +16,11 @@
     "`pip install jupyter`\n",
     "\n",
     "Then to run it, simply enter `jupyter-notebook` into the Raspberry Pi terminal\n",
-    "and then navigate the `picamera2/examples/notebooks` folder to find this notebook.\n"
+    "and then navigate the `picamera2/examples/notebooks` folder to find this notebook.\n",
+    "\n",
+    "Note: If running in chromium, you may see better performance by turning off hardware acceleration in the chromium settings.\n",
+    "\n",
+    "***"
    ]
   },
   {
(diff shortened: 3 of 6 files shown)


IanTBlack:
I've gone through the style checker a couple times now. It seems to lag a little with changes. I think I'm as up to date as I can be. I'll still need some feed back on what to do about the doc string. I can clean it up in module or start a readme.

davidplowman:
Hi, thanks for submitting this. I was wondering if we might tidy some things up before I look further at it.

1. Would it be possible to fix the style checker issues? It seemed like there are still enough fairly basic complaints in the output. I'm happy to ignore the style checker when it really doesn't help, but I think we can still do a little more to make it happy here.

2. All the automatic tests seem to be failing. Did you run `tools/run_tests.py` and did that report everything as passing? Could it be that some more Python modules need to be installed which are missing on the test machine? If so, please let me know what those are. The installation instructions in the REAME would also need updating in that case.

3. I haven't really looked at the functionality yet. One thing I did notice were some replacements of `picam2.start(controls)` by `picam2.start()` follwed by `picam2.set_controls(controls)`. These are not the same thing so that will need to be put back. Unfortunately it's a slightly tricky thing to create a test for, but maybe I should try. One thing I quite like about using dicts for controls is that it's easy to pass them around, store them, apply them later, even merge them with the `|` operator. Is that something that is still possible here (I haven't looked deeply enough yet)? But I certainly agree that some enhancements providing such things like name-checking, type-checking etc. would be welcome.

4. I'd like to see a somewhat tidier git commit history if that's possible. I'd like to avoid commits that say "fix a previous commit" - I'd rather just the two were merged together. Ditto for reverting a previous commit in the set - can you just rebase and remove the commit. Also where we fix style checker issues, if possible I'd prefer to update the original commit. I could probably be clearer about this in the guidance in the README, I'll try and expand that a bit when I'm next editing it.

Thanks again!

IanTBlack:
Hi David,

I can work on tidying this up.

1) The changes that the style checker recommended such as missing whitespace and spacing definitely exist. Not sure what to do about W503 or W504 conflicts because it raises an error at either one.

2) There is a conflict with del. I unfortunately don't see these with PyCharm. PyCharm also auto imports a lot of things so I will need to add some imports.

3) Set controls was reworked to work before and after initial start. You can still input things as dicts, but instead of passing it to start, you must define the controls before or after.  You can see methods for how to use it in the Jupyter Notebook. Passing them to the prebuilt configurations still works too.

4) I will close and rework.

- Ian

IanTBlack:
Hi David,

After the adding the _USER variable to run tests I was able to get this PR to pass all tests except for the OpenCV ones. 
Since OpenCV is not a requirement of picamera2, can those examples be removed?
I also don't see a reason to keep the context and preview tests in the tests folders, since those were originally intended to show functionality. I can replace those with notebooks later.

-Ian

IanTBlack:
And by remove, I mean removed from test_list.txt. Not actually removed from the repo.

PR #87, opened 2022-05-02
Fix for systems that no longer have /home/pi.

IanTBlack:
pi user is no longer the default option on newer images.
Adds _USER variable to determine current user.


IanTBlack:
There appears to be a conflict with the smoke tests yaml when github runs its checks, but I have confirmed that this doesn't cause any problems when I use run_tests.py on my own RPi.

davidplowman:
Yes, the environment is a bit different when things run for GitHub, I think. But the automated tests do really need to work. Could we set the _USER variable in a try block and default it to pi if there's an error? Also, please don't forget to sign your commit. Thanks!

IanTBlack:
Sure thing.

PR #88, opened 2022-05-02
Added controls module.

IanTBlack:
### Is this a major or minor change?
Major. Suggest adding to new branch for testing.

### What is the purpose of this PR?
This PR is a rework of PR #86.
Adds a controls module that provides more options for assigning a control to the camera.
Please see [control_example.ipynb](https://github.com/IanTBlack/picamera2/blob/controls/examples/notebooks/controls_example.ipynb).

### If necessary, were all examples updated?
All examples were updated to use new control scheme.

### Does this PR pass run_tests.py?
This PR passes all tests with the exception of the OpenCV examples. 
OpenCV is not a requirement of picamera2. Recommend removing from test list or adding a checker for OpenCV.

### Was style-checking performed?
Style checking was done via PyCharm in accordance with PEP8.
This also updated formatting not associated with this PR.

Signed-off by: Ian Black <blackia@oregonstate.edu>


.gitignore
```diff
@@ -4,3 +4,6 @@ __pycache__
 .spyproject
 /.idea
 .idea
+*.jpg
+*.png
+*.h264
```
davidplowman, line 9: Could this be a separate pull request, please? I know it's a bit annoying, but it's not really related to the rest of this pull request, could be merged earlier, and would leave less stuff to review here. Thanks!

examples/capture_image_full_res.py
```diff
@@ -17,7 +17,6 @@
 image = picam2.switch_mode_and_capture_image(capture_config)
 image.show()
 
-
```
davidplowman, line 20: Could this whole commit be a separate pull request too, please? Sorry that this is a bit of a hassle, but it's not really related to the other changes here, could be merged earlier, and would leave less in this PR to review. Thanks!

picamera2/controls.py
```diff
@@ -156,5 +156,5 @@ def defaults(self) -> dict:
 
     @property
     def reset(self) -> None:
-        for ctrl, val in self.defaults().items():
+        for ctrl, val in self.defaults.items():
```
davidplowman, line 159: Could this be squashed into the original commit that introduced this code? Again, it just makes the review job a little bit easier. Thanks!

picamera2/picamera2.py
```diff
@@ -235,6 +235,7 @@ def preview_configuration(self, main={}, lores=None, raw=None, transform=libcame
         lores = self.make_initial_stream_config({"format": "YUV420", "size": main["size"]}, lores)
         raw = self.make_initial_stream_config({"format": self.sensor_format, "size": main["size"]}, raw)
         controls = {"NoiseReductionMode": 3} | controls
+        self.set_controls(controls)
```
davidplowman, line 253: I think this might be a problem. The intention is that while the camera is running you can create another configuration (for example, call still_configuration() while the preview is running). I'm guessing this might cause the control values intended for the new mode to be sent during the preview?

examples/exposure_fixed.py
```diff
@@ -10,6 +10,6 @@
 
 preview_config = picam2.preview_configuration()
 picam2.configure(preview_config)
-
-picam2.start({"ExposureTime": 10000, "AnalogueGain": 1.0})
+picam2.set_controls({"ExposureTime": 10000, "AnalogueGain": 1.0})
+picam2.start()
```
davidplowman, line 14: It looks to me like fixing up the examples here leaves quite a string of commits where the test set is broken, is that right? I'm wondering if we should really squash this with the commit that changed the start function signature.

Though I'm still unsure about the change to the start function. I'm quite keen that all the state associated with a configuration is "obvious", and there isn't anything hidden in the Picamera2 object. The idea being that if you call start the with the same configuration and parameters, you're guaranteed the same outcome.

examples/capture_image_full_res.py
IanTBlack: This was part of style cleanup associated with PEP8 that was run on everything, not just my additions. I can revert that commit and focus on only my stuff if you would like.

picamera2/controls.py
IanTBlack: Can do. Just need some input on other things before I force push some changes.

picamera2/picamera2.py
IanTBlack: That would be correct. There are two options I think could work.

1) Remove set_controls from prebuilt configurations and move it to .configure so the controls aren't changed until the user submits configure.
2) Remove the controls options from the prebuilts. Separate control setting from stream configuration. I don't think this would difficult to conceptualize with the right documentation.

examples/capture_image_full_res.py
davidplowman: Yes please, when there's a long list of commits to review it's helpful if other stuff doesn't creep in, though I know how tempting that always is!

picamera2/picamera2.py
davidplowman: I think 1 sounds OK. There are specific reasons why particular configurations "recommend" particular control values, and I wouldn't want to do anything that makes it easier for users to end up with inappropriate choices by forgetting to do something.

examples/notebooks/controls_example.ipynb
```diff
@@ -0,0 +1,660 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "my_controls = {'NoiseReductionMode': 0,\n",
```
davidplowman, line 86: All the "enum" type controls will want changing to use the Python enums provided by libcamera, e.g. `libcamera.NoiseReductionMode.Off`. I'd like to deprecate the use of "magic ints" where we can.

picamera2/picamera2.py
```diff
@@ -19,6 +19,7 @@
 from enum import Enum
 import piexif
 
+from picamera2.controls import Controls
```
davidplowman, line 22: I know I like to separate my commits nicely, but I think this one could be combined with the first commit that uses the new controls!!

IanTBlack: I started to think that after the first few comments. Just did that!

picamera2/picamera2.py
```diff
@@ -522,26 +522,24 @@ def stream_configuration(self, name="main"):
         """Return the stream configuration for the named stream."""
         return self.camera_config[name]
 
-    def start_(self, controls={}):
-        """Start the camera system running."""
-        if self.camera_config is None:
-            raise RuntimeError("Camera has not been configured")
-        if self.started:
-            raise RuntimeError("Camera already started")
-        if self.camera.start(self.camera_config["controls"] | controls) >= 0:
```
davidplowman, line 516: I think I'd prefer these error checks in `start_` rather than in `start`. There's a convention I've been trying to stick to (obviously, there's a bit of a problem in that no one else knows this, so sorry about that) where functions that end in `_` are guaranteed to be safe to call from the camera event loop, and the non-`_` versions aren't. Now, `start` and `start_` (and `configure` and `configure_`) are a bit different in that the camera can't be running, so they're actually the same. I was in several minds whether to create the two versions of each or not, but given that I did, I think that on balance I would prefer the old version. Thanks!

examples/notebooks/controls_example.ipynb
IanTBlack: I don't see any problems with sticking keeping integers as another method for denoting a mode, as long as the documentation is robust.  I do think using strings is a better method because it clearly represents the mode while reducing the complexity in entering the value.  Requiring the passing of libcamera enums could be confusing, `libcamera.NoiseReductionMode.HighQuality` is a lot to type for something that could also be entered as a string or integer. It requires users to have some knowledge of libcamera, meaning that they would potentially have to review documentation that is separate from picamera2. This could be mitigated by shortening the libcamera enums in the picamera2 module and by providing the right documentation when picamera2 is released.

I understand that picamera2 and libcamera are tightly coupled, but I see picamera2 being used in more educational settings than libcamera, so having options for both general and advanced users is nice to have.

picamera2/controls.py
```diff
@@ -0,0 +1,160 @@
...
+    FrameDurationLimits: (tuple, list) = None
+    MaxLatency: int = None
+    NoiseReductionMode: (int, str) = None
+    PipelineDepth: int = None
+    Saturation: (float, int) = None
+    ScalerCrop: (tuple, list) = None
+    SceneFlicker: (int, str) = None
+    Sharpness: (float, int) = None
+    TestPatternMode: (int, str) = None
+
```
davidplowman, line 44: One thing that's bothering me a bit is that we have a hard-coded list of stuff here, and how we are going to maintain it. I'm thinking of libcamera adding new controls, or changing the definitions of existing controls. Will they work immediately (as they do now) or will we have to update Picamera2 to "catch up"? And who does the "catching up" and how quickly will they do it? Could there be a way of supporting "unfamiliar" controls?

picamera2/controls.py
```diff
@@ -0,0 +1,160 @@
...
+        """This function returns the possible ranges for each controls."""
+        _ranges = {}
+        for ctrl, val_tuple in self._camera_controls.items():
+            _ranges[ctrl] = {'min': val_tuple[0], 'max': val_tuple[1]}
+        _ranges = dict(sorted(_ranges.items()))
+        return _ranges
+
+    @property
+    def defaults(self) -> dict:
+        """This function takes the default camera controls and displays them
```
davidplowman, line 137: This one makes me a bit nervous. I'm worried that users might think that `'picam2.set_controls(picam2.controls.defaults))` is a good idea... it resets all the camera controls to its normal defaults, right? Only it really isn't!

picamera2/controls.py
```diff
@@ -0,0 +1,160 @@
...
+        This function _blanks all controls to None.
+        Its only use is to clear the controls at camera close. This is so
+        that if a user decides to reopen the camera after close, the old
+        control settings do not persist and carry over to the next start.
+        """
+        for attr in self.__annotations__.keys():
+            setattr(self, attr, None)
+
+    @property
+    def fps(self) -> int:
```
davidplowman, line 105: Framerates should be floating point numbers.

picamera2/controls.py
IanTBlack: Would the proper method to reset to defaults then be to stop, close and then restart?

picamera2/controls.py
```diff
@@ -0,0 +1,160 @@
...
+    def current(self) -> dict:
+        """
+        This function returns the current user-defined settings of the camera.
+        It is only updated through the CompletedRequests module.
+        """
+        return self._current
+
+    @property
+    def _blank(self) -> None:
+        """
```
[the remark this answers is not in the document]
IanTBlack: Yes, removing it as a property would probably be better. I would avoid the use of clear in case of confusion with base Python functionality.

picamera2/controls.py
IanTBlack: I'm also bothered by this. If there is a yaml file on the libcamera end that can be parsed, then that would be a solution. In theory, they would continue to work as long as we know what the name is and what types it can be.

picamera2/controls.py
IanTBlack: I agree. It was a hopeful addition.

examples/notebooks/controls_example.ipynb
davidplowman: It's fine that "magic ints" work, and we should say that "you can use ints too if you want". I just wouldn't want them to be the first thing that people see, or to become the preferred way of coding. I think both enums and strings are fine, it's just that if I saw a piece of code using ints, I honestly wouldn't have a clue what was going on!

examples/exposure_fixed.py
davidplowman: I think I'm probably OK with using the set_controls function instead of passing controls in start, I'm just struggling a bit to get used to the idea. I think I remain nervous about separating stream configuration and controls any more, because I think there are particular use cases where you want to specify appropriate defaults both for the configuration and for specific controls.

picamera2/picamera2.py
IanTBlack: Roger. Makes sense. To me, the '_' versions are for internal use only, so they should have functionality that other internal processes use. I'll switch it back. 

I would like to discuss the possibility of switching it to a preceding '_' so there is no confusion when someone is using an interactive IDE or if they are using the module as a source of information. That can be another day for another PR though.

picamera2/controls.py
IanTBlack: Just to add to this, what would the effort be like to add all of the camera controls to the Picamera2().camera.controls output? Presumably that comes from a place libcamera, so that would also be source for the current camera settings and would reduce maintenance effort in the future..

picamera2/picamera2.py
davidplowman: I think `_` for purely internal things sounds like a good idea. Sometimes it takes a while to learn exactly what really needs to be exposed and what doesn't, of course. But I agree it's something for another PR, my brain can only take so much at one time!

davidplowman:
Hi, I'd like to suggest a way to take this patch set forward. As you can probably tell, large patch sets can be quite difficult to digest and review, especially if they haven't been discussed before. So perhaps we can break this down and tackle things in stages.

First of all, we've talked a little bit about that API change to `Picamera2.start`. I think I can agree that it might send an easier message to users if we remove the controls parameter. So perhaps we could start with a PR for that? I would suggest updating the API, and calling `set_controls` at the end of `configure`. We can also update the examples in that commit so that everything still passes the tests.

After that we can move on and discuss proposals for the implementation that goes underneath this, and what benefits we will get by changing it. Thanks!

IanTBlack:
I'll see if I can find some time today and tomorrow to make those changes. 

I've also been thinking about maintenance and a method to obtain a list of controls. There appear to be two methods for getting control names from libcamera. 

`libcamera.CameraManager.singleton().cameras[idx].controls` where `idx` is the camera number. 

This appears to give you only some of controls.

When I do `dir(libcamera)` or `dir(libcamera._libcamera)`, I also see some control options that aren't in the first method.


If the first method to get the controls for the camera were updated with all of the controls, it would would not be difficult to parse this and then do checks against it. This would particularly be useful if a control becomes camera specific and as new controls are added. I think then the only thing someone would have to do is maintain this list with respect to the python bindings for libcamera.

It is currently outside my scope of knowledge to make adjustments to libcamera.  Is there a method for functionality recommendation? I know I am asking a lot here, but it would also be beneficial if the list of controls could be expanded to include something like this.

`{'ScalerCrop': { 'valid_type': tuple, 'length': 4, 'default' : 0, 'min':  0 , 'max':  65535}, ....}`

This would be helpful in that checks could be written in Python that prevent the user from inputting the wrong values. Presumably libcamera would already handle that. I suppose you would just need to decide if you are okay with libcamera handling and displaying errors vs picamera2. I think if you had picamera2 handle errors then it would be easier for the end user to troubleshoot. Particularly those less experienced.

What I can do to avoid hardcoding the controls names is to read in this list of controls and then dynamically create class variables based on those names. Does that sound like a good path?

IanTBlack:
@davidplowman I'm going to close this PR as well. Seems like there are a lot of tiny editions in other PRs that may be implemented soon, so I will wait a day or two until those complete so I don't create any conflicts. If you could let me know what  you think about listing all of the controls through the libcamera `.controls` path, that would be appreciated. In the mean time, I'll work on what you suggested with the start function.

PR #89, opened 2022-05-03
Update contributing guidelines to be a bit clearer

davidplowman:


davidplowman:
Surprised to see 4 failures here, but they all look like the "bogus" one which we're working on down in libcamera itself.

PR #90, opened 2022-05-03
Preview merge

davidplowman:
The QtPreview and QtGlPreview classes are virtually identical, so this pair of commits tidies up some minor differences and then merges the two together. There should be no functional difference, just slightly less code to maintain.


chrisruk:
@davidplowman Looks good, just a small question would it be better to put the 'imports' that are in one of the functions, at the start of the file instead?

davidplowman:
Yes, good point, this is all quite ugly. The problem we have is that Qt is really fussy, and if you ever call Qt functions from more than one thread then it complains, and sometimes it even dies horribly. Unfortunately, even _importing_ Qt stuff seems to qualify, so this too has to happen in the thread that is going to be running the Qt event loop.

The root cause is that Qt really expects to be the main thread, and so in this sense we are using it incorrectly. But I'm not aware of any better solutions...

chrisruk:
Gotcha, thanks for explaining that, will merge now

PR #92, opened 2022-05-04
Initial attempt to set gop size

chrisruk:
@davidplowman Hi, i'm wondering if you can see any issues with this, currently I get an 'invalid argument' when attempting to set the gop size with this patch.  

But it can set the 'repeat' flag ok, so I'm a bit stuck as to what could be wrong with respect to the gop size flag.


picamera2/encoders/h264_encoder.py
```diff
@@ -62,16 +63,21 @@ def _start(self):
         fmt.fmt.pix_mp.plane_fmt[0].sizeimage = 512 << 10
         fcntl.ioctl(self.vd, VIDIOC_S_FMT, fmt)
 
+        controls = [(V4L2_CID_MPEG_VIDEO_GOP_SIZE, self._gopsize)]
```
davidplowman, line 66: I have a sneaky suspicion this should be V4L2_CID_MPEG_VIDEO_H264_I_PERIOD. What a fabulous interface, it's as though someone designed it to trip people up! That control parameter probably wants adding to our v4l2.py, I tried it and it seemed like it wasn't there.

chrisruk: Thanks, that did the trick :) I've just added that flag to the v4l2 package and pushed to pypi too.

Do you think I should rename gopsize to iframeperiod possibly?

I ran v4l2-ctl --all --device /dev/video11

And it seems to be listed as i frame period in there too.

```
            h264_i_frame_period 0x00990a66 (int)    : min=0 max=2147483647 step=1 default=60 value=60
```

davidplowman:
Don't really mind, both names are fine with me!

Do you want to squash those two commits together and update the commit message? We don't really want a commit lurking in the tree where stuff is actually broken, if we can avoid it.

Once that's all ready, I'm good to merge. Thanks!

PR #93, opened 2022-05-04
Dng support

davidplowman:
There are two commits here.

The first adds support for saving DNG files using the PiDNG library.

The second just adds some tests and updates the documentation. The installation instruction now include installing PiDNG.


davidplowman:
@schoolpost I've squashed your original commits together to make a single "tidy" commit, and also added your "signature" to confirm that you're happy to donate the work in that commit to us. Please let me know that you're OK with that! On reflection I've also made it mandatory to install PiDNG, I think we're early enough in the project to force everyone to install all the correct stuff.

You'll notice I had a couple of issues and have sent you a couple of pull requests for PiDNG. The first relates to the raw image buffer sizes for the v1 and v2 cameras. The second fixes the colour matrix reported in the DNG file which should make them look a bit better.

(The test failures are nothing to worry about at this point. The automatic test machine doesn't have PiDNG installed yet so everything fails! I'll get round to that, only it's running a v2 camera so actually the DNG tests will still fail until we've fixed the image size issue.)

Thanks for all your work on this!

PR #94, opened 2022-05-05
Use Black for Formatting

meawoppl:
A quick demo of how to make `black` auto-format everything. This also shifts format checking to gh infrastructure.

All I did to generate a bulk of the changes in this PR was:
`black picamera2` which in turn output the reformatted code:
image

This has the side benefit of killing a lot of diff noise between PR's

Signed-off-by: Matthew Goodman <matt@exclosure.io>

### DIFF
--- a/picamera2/converters.py
+++ b/picamera2/converters.py
@@ -1,9 +1,13 @@
 import numpy as np
 
 
-YUV2RGB_JPEG      = np.array([[1.0,   1.0,   1.0  ], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
-YUV2RGB_SMPTE170M = np.array([[1.164, 1.164, 1.164], [0.0, -0.392, 2.017], [1.596, -0.813, 0.0]])
-YUV2RGB_REC709    = np.array([[1.164, 1.164, 1.164], [0.0, -0.213, 2.112], [1.793, -0.533, 0.0]])
+YUV2RGB_JPEG = np.array([[1.0, 1.0, 1.0], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
+YUV2RGB_SMPTE170M = np.array(
+    [[1.164, 1.164, 1.164], [0.0, -0.392, 2.017], [1.596, -0.813, 0.0]]
+)
+YUV2RGB_REC709 = np.array(
+    [[1.164, 1.164, 1.164], [0.0, -0.213, 2.112], [1.793, -0.533, 0.0]]
+)
 
 
 def YUV420_to_RGB(YUV_in, size, matrix=YUV2RGB_JPEG, rb_swap=True, final_width=0):
@@ -19,8 +23,8 @@ def YUV420_to_RGB(YUV_in, size, matrix=YUV2RGB_JPEG, rb_swap=True, final_width=0
 
     YUV = np.empty((h2, w2, 3), dtype=int)
     YUV[:, :, 0] = YUV_in[:n].reshape(h, w)[0::2, 0::2]
-    YUV[:, :, 1] = YUV_in[n:n + n4].reshape(h2, w2) - 128.0
-    YUV[:, :, 2] = YUV_in[n + n4:n + n2].reshape(h2, w2) - 128.0
+    YUV[:, :, 1] = YUV_in[n : n + n4].reshape(h2, w2) - 128.0
+    YUV[:, :, 2] = YUV_in[n + n4 : n + n2].reshape(h2, w2) - 128.0
 
     if rb_swap:
         matrix = matrix[:, [2, 1, 0]]
--- a/picamera2/encoders/jpeg_encoder.py
+++ b/picamera2/encoders/jpeg_encoder.py
@@ -4,11 +4,13 @@ from picamera2.encoders.multi_encoder import MultiEncoder
 
 
 class JpegEncoder(MultiEncoder):
-    def __init__(self, num_threads=4, q=85, colour_space='RGBX'):
+    def __init__(self, num_threads=4, q=85, colour_space="RGBX"):
         super().__init__(num_threads=num_threads)
         self.q = q
         self.colour_space = colour_space
 
     def encode_func(self, request, name):
         array = request.make_array(name)
-        return simplejpeg.encode_jpeg(array, quality=self.q, colorspace=self.colour_space)
+        return simplejpeg.encode_jpeg(
+            array, quality=self.q, colorspace=self.colour_space
+        )
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -26,6 +26,7 @@ VIEWFINDER = libcamera.StreamRole.Viewfinder
 
 class Preview(Enum):
     """Enum that applications can pass to the start_preview method."""
+
     NULL = 0
     DRM = 1
     QT = 2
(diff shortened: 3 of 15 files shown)


meawoppl:
Note that style check now takes about... 4 seconds:
image

davidplowman:
Hi, thanks for posting this it looks interesting. Can you say how I could enforce this kind of style during development, for example by using the git pre-commit hook? Also, have you changed the style checking process on the test machine to complain if people don't stick to this style or not, I didn't quite understand that.

meawoppl:
There are a number of options for this:
- Before commiting run `black picamera2` from the repository root
- Install black and add `black picamera2` to the `.git/hooks/pre-commit`
- Install and use the `pre-commit` [framework](https://pre-commit.com/) (which is what I do in my projects), which will manage installation of many potential pre-commit options (including black). It's nice because it comes with a raft of options that you can scale in as the code becomes more mature. Starting with formatting is easy, then adding lint, doc requirements, type-checking etc all just becomes a couple lines of config

Here is what that looks like in a project I am working on:
precommit-demo

meawoppl:
Also, this PR is likely to be conflicted by basically any other change, so let me know when/if you want it and I can freshen it :)

meawoppl:
And to answer your earlier question. [This line](https://github.com/raspberrypi/picamera2/pull/94/files#diff-6001a1d79f0d7c8dd0eb053d986e182e24ad87c11f8daf3d6428da2e1e36fbe0R19) will cause the CI to reject PR's that are not consistant with the formatting requirements.

meawoppl:
As a totally incidental side benefit, this takes the style check responsibility off the pi you are using to run tests :)

naushir:
This change seems reasonable, but there are a couple of things that I can see we need to address:

- The github action command probably wants to be ``black --check --diff picamera2`` to show what formatting changes would be needed.
- There does not seem to be a way to run black on a diff/patch as far as I can tell.  This is not ideal, since if we chose to ignore a particular style dictated by the tool for a single PR, the tool will end up complaining about it for all subsequent PRs, which defeats the purpose of the pre-commit check.

meawoppl:
Hmmm. The flags are easy to switch out. Let me start there.

Following up on the second bullet point. Looking through the checker it seems that it completely checks all files/lines, then filters to lines contained in the diff only. I am more than a little confused as to how that can deal with multi-line expressions in a sane way, but that isn't the question here. `black` is named for the Henry Ford quote, its kinda hostile by design.
image

It might just not meet the needs of this group if y'all want to have exceptionally formatted sections.

FWIW, the not having redo PR's for code style issues is about 50% of the appeal, another 40% is not having to educate folks that contribute about the particular flavor of code style. :man_shrugging:

naushir:
Apparently running on a diff has been a feature request for some time now - https://github.com/psf/black/issues/830.  Looks unlikely that this will be put into mainline.

There are a few wrapper-of-sorts that does seem to do this - https://github.com/akaihola/darker and https://github.com/wbolster/black-macchiato.  Perhaps it's worth looking to see if those could be used instead?

picamera2/converters.py
```diff
@@ -1,9 +1,13 @@
...
-YUV2RGB_JPEG      = np.array([[1.0,   1.0,   1.0  ], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
-YUV2RGB_SMPTE170M = np.array([[1.164, 1.164, 1.164], [0.0, -0.392, 2.017], [1.596, -0.813, 0.0]])
-YUV2RGB_REC709    = np.array([[1.164, 1.164, 1.164], [0.0, -0.213, 2.112], [1.793, -0.533, 0.0]])
+YUV2RGB_JPEG = np.array([[1.0, 1.0, 1.0], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
+YUV2RGB_SMPTE170M = np.array(
+    [[1.164, 1.164, 1.164], [0.0, -0.392, 2.017], [1.596, -0.813, 0.0]]
+)
+YUV2RGB_REC709 = np.array(
+    [[1.164, 1.164, 1.164], [0.0, -0.213, 2.112], [1.793, -0.533, 0.0]]
+)
```
naushir, line 11: I think I prefer the original formatting.  Or possibly all 3 matrices to look  something like:

```
YUV2RGB_REC709 = np.array(
    [ [1.164, 1.164, 1.164],
      [0.0, -0.213, 2.112],
      [1.793, -0.533, 0.0] ]
)
```

picamera2/encoders/jpeg_encoder.py
```diff
@@ -4,11 +4,13 @@
...
         super().__init__(num_threads=num_threads)
         self.q = q
         self.colour_space = colour_space
 
     def encode_func(self, request, name):
         array = request.make_array(name)
-        return simplejpeg.encode_jpeg(array, quality=self.q, colorspace=self.colour_space)
+        return simplejpeg.encode_jpeg(
+            array, quality=self.q, colorspace=self.colour_space
+        )
```
naushir, line 16: Not sure how I feel about the formatting here and in all the other places where the function arguments are put on the next line.  The original reads easier to me.

picamera2/picamera2.py
```diff
@@ -107,13 +110,19 @@ def _reset_flags(self) -> None:
         self._encoder = None
         self.request_callback = None
         self.completed_requests = []
-        self.lock = threading.Lock()  # protects the functions and completed_requests fields
+        self.lock = (
+            threading.Lock()
+        )  # protects the functions and completed_requests fields
         self.have_event_loop = False
```
naushir, line 116: The original looks better and more concise in the original.

picamera2/picamera2.py
```diff
@@ -341,7 +421,9 @@ def check_camera_config(self, camera_config: dict) -> None:
                 raise RuntimeError(f"'{name}' key expected in camera configuration")
 
         # Check the entire camera configuration for errors.
-        if not isinstance(camera_config["colour_space"], libcamera._libcamera.ColorSpace):
+        if not isinstance(
+            camera_config["colour_space"], libcamera._libcamera.ColorSpace
+        ):
             raise RuntimeError("Colour space has incorrect type")
```
naushir, line 427: Prefer the original.

picamera2/picamera2.py
```diff
@@ -411,21 +505,46 @@ def align_stream(self, stream_config: dict) -> None:
...
+        return fmt in (
+            "NV21",
+            "NV12",
+            "YUV420",
+            "YVU420",
+            "YVYU",
+            "YUYV",
+            "UYVY",
+            "VYUY",
+        )
```
naushir, line 518: I think the original single line version looks better.

naushir (commented):
Here are some comments on the formatting produced by Black.  Of course, these preferences are entirely personal to my tastes.

picamera2/picamera2.py
meawoppl: I think it was just trying to maintain the comment on the same line here...

picamera2/picamera2.py
meawoppl: Concur.

picamera2/previews/q_gl_picamera2.py
```diff
@@ -50,12 +51,18 @@ def choose_config(self):
...
+            EGL_GREEN_SIZE,
+            8,
+            EGL_BLUE_SIZE,
+            8,
+            EGL_ALPHA_SIZE,
+            0,
+            EGL_RENDERABLE_TYPE,
+            EGL_OPENGL_ES2_BIT,
             EGL_NONE,
         ]
```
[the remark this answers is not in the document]
meawoppl: Oh yeah, that is kinda and uglee tranform. Maybe these want to be a `dict` then tranformed at the call site?

meawoppl:
Seems like this ins't a real improvement here.

PR #95, opened 2022-05-05
Kill `import *` instances

meawoppl:
This PR does two things:
- removes a majority of star imports from codebase. For a lengthy set of examples for why this is a good idea, checkout this [SO post](https://stackoverflow.com/questions/2386714/why-is-import-bad). 
- Adds stdout and stderr reporting from tests.

This leaves star imports in a few places:
- Submodules that import a slew of V4L2 symbols
- Submodules that used the OpenGL.**.* structures
These seem pretty hairy, but well contained.

Incidentally, this helped find/discover a number of places where `np`, `time`, and `os` were imported by proxy and thusly fragile.

### DIFF
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -50,6 +50,21 @@ def clean_directory():
         os.remove(os.path.join(cwd, f))
 
 
+def indent(text, num_spaces=4):
+    between = "\n" + " " * num_spaces
+    return between.join(text.splitlines()) + "\n"
+
+
+def print_subprocess_output(exc: subprocess.CalledProcessError):
+    if exc.stdout is not None:
+        print("=" * 20 + "    STDOUT    " + "=" * 20)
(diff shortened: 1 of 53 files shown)


tools/run_tests.py
```diff
@@ -50,6 +50,21 @@ def clean_directory():
...
+
+def print_subprocess_output(exc: subprocess.CalledProcessError):
+    if exc.stdout is not None:
+        print("=" * 20 + "    STDOUT    " + "=" * 20)
+        print(indent(exc.stdout.decode('utf-8')))
+
+    if exc.stderr is not None:
+        print("=" * 20 + "    STDERR    " + "=" * 20)
+        print(indent(exc.stderr.decode('utf-8')))
+
```
meawoppl, line 66: @davidplowman this is the only substantive change other than the imports.

PR #97, opened 2022-05-06
Update to start and start_

IanTBlack:
- Removed the controls option from the start function. I will propose a future PR with enhancement of controls.
- Moved event_loop to dissuade confusion for beginners.
- Updated doc strings.
- Updated log and error messages with camera index.
- Updated relevant examples to proposed changes.

The preview cycle test could be reworked. If there is a PR that reworks tests in the future, I would recommend that test be split into individual tests so that failures can be rapidly identified.

Signed-off By: Ian Black <blackia@oregonstate.edu>

### DIFF
--- a/examples/exposure_fixed.py
+++ b/examples/exposure_fixed.py
@@ -8,8 +8,8 @@ import time
 picam2 = Picamera2()
 picam2.start_preview(Preview.QTGL)
 
-preview_config = picam2.preview_configuration()
+preview_config = picam2.preview_configuration(controls = {"ExposureTime": 10000,
+                                                          "AnalogueGain": 1.0} )
 picam2.configure(preview_config)
-
-picam2.start({"ExposureTime": 10000, "AnalogueGain": 1.0})
+picam2.start()
 time.sleep(5)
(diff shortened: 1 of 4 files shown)


davidplowman:
Hi, thanks for sending this. Mostly it looks good to me, just a few small points.

1. I was expecting us to do a `self.set_controls(self.camera_config["controls"])` (or equivalent) at the end of `Picamera2.configure`, as that would make it easier for a user subsequently to overwrite the "recommended" control values that are in there. Does that sound right?
2. I'm wondering about squashing all these commits together. Generally I don't like doing that, but I possibly dislike having commits where the automatic tests are broken even less. What do you think?
3. Could you fix the style checker complaint too, please.
4. I also generally dislike mixing cosmetic and functional changes together, because the former make the latter harder to review. On this occasion, though, I guess they're small enough that I'm not too bothered.

Thanks!

davidplowman:
Hi Ian, just wondering how you wanted to progress with this. The reason I'm asking is that there's been a huge amount of tidying up and quite a few new features over the last week or so, and we've even got the "official" libcamera Python bindings now. We're wanting to get all that merged which will then require a simultaneous update to both libcamera and Picamera2.

After that I'm keen to "re-advertise" the existence of Picamera2 again on the forum. At that point I'd like to start "raising the bar a little" on breaking API changes. I think we're still at a point where I'm happy to change them if there are good reasons, and obviously enhancements (including improved or new APIs) are welcome. So this does mean that we might want to get this change made sooner rather than later.

I'm happy to go with whatever you would prefer to do - do you want to carry on and get this PR merged, or alternatively just let me know and I'll make this API change myself. Then we can see where the topic of controls takes us after that...  Thanks!

PR #104, opened 2022-05-10
New FfmpegOutput class allows writing of mp4 files and audio streams

davidplowman:
This pair of commits adds a new `FfmpegOutput` class. It allows the encoded video stream to be sent to _FFmpeg_ for output, allowing it to be muxed into various different types of container, including _mp4_ files. You can even get _FFmpeg_ to record an audio stream into the output file.

This is perhaps a little bit experimental in that there's so much stuff that you could do with _FFmpeg_, it's hard to know where to draw the line as to what folks should be able to try and what you might reasonably expect to work. I also know of no easy way to get precise timestamps (that _Picamera2_ has) into _FFmpeg_, so I'm having to let it resample those which will doubtless incur some jitter.

Still, on balance it seems to work pretty well, and _mp4_ files and audio are a good feature to have.

The first commit implements the new class, the second adds examples and updates documentation. Note that there's a new dependency to install (`sudo apt install -y python3-prctl`). Can we assume _FFmpeg_ is installed on all Pis? That's also something to check, though it would only fall over if you try to use the class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Finally fetch the *Picamera2* repository. There are some dependencies to downloa
 ```
 cd
 sudo pip3 install pyopengl piexif simplejpeg PiDNG
-sudo apt install -y python3-pyqt5 python3-numpy
+sudo apt install -y python3-pyqt5 python3-numpy python3-prctl ffmpeg
 git clone https://github.com/raspberrypi/picamera2.git
 ```
 
@@ -40,6 +40,12 @@ To make everything run, you will also have to set your `PYTHONPATH` environment 
 ```
 export PYTHONPATH=/home/pi/picamera2
 ```
+
+#### FFmpeg
+
+We have suggested installing _FFmpeg_ in order to use _Picamera2_. It is, however, only required to support certain complex file types, such as writing output to _mp4_ files. If you do not require this functionality then _FFmpeg_ is not required, and you should avoid trying to use the `FfmpegOutput` class.
+
+We note that it is installed by default in Raspberry Pi OS, but not in Raspberry Pi OS Lite.
 
 #### OpenCV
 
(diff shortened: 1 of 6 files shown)


chrisruk:
This looks really cool, couple of points

Occasionally I've been getting this syntax issue, which I just got now, probably unrelated to this pull request though:
```
/tmp/picamera2/picamera2/picamera2.py:701: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if name is "raw" and self.is_Bayer(self.camera_config["raw"]["format"]):
```

I guess we need to add 'python-prctl' to the install_requires for setup.py?  (I did have to do: sudo apt-get install build-essential libcap-dev, for the pip install of python-prctl to succeed)

ffmpeg didn't seem to be installed by default for me (as I used the lite image).

davidplowman:
> This looks really cool, couple of points
> 
> Occasionally I've been getting this syntax issue, which I just got now, probably unrelated to this pull request though:
> 
> ```
> /tmp/picamera2/picamera2/picamera2.py:701: SyntaxWarning: "is" with a literal. Did you mean "=="?
>   if name is "raw" and self.is_Bayer(self.camera_config["raw"]["format"]):
> ```

Yes, I've seen that too, don't really know why we weren't seeing it earlier as it doesn't seem even remotely related to this! But I think a small PR is in order.

> 
> I guess we need to add 'python-prctl' to the install_requires for setup.py? (I did have to do: sudo apt-get install build-essential libcap-dev, for the pip install of python-prctl to succeed)

Indeed, installing through pip is a bit annoying for that reason. In the installation instructions I've suggested doing it through apt which (as far as I can tell) "just works". Would that save us having to update setup.py?

> 
> ffmpeg didn't seem to be installed by default for me (as I used the lite image).

Ah, OK. That probably wants a doc update. Presumably `sudo apt install -y ffmpeg` did the trick for you? Let me add that and re-push.

chrisruk:
Yep that's what I did to install ffmpeg.

I just checked if you add 'python-prctl' to setup.py, and install from apt, the pip3 install . --user would use the apt version fine.  So it might be worth adding to setup.py, to make sure it is required?

davidplowman:
Sure, do you want to do a little PR for that once this one is merged? I'll do one for that syntax complaint thing, and re-post this one with a doc update for ffmpeg.

chrisruk:
Sure, yep I'll create a pull request for setup.py

PR #107, opened 2022-05-10
Improve yuv_to_rgb.py example

davidplowman:
When OpenCV is available, it has more efficient YUV420 to RGB
conversion functions which you may be able to use. Documentation for
this example also updated.


chrisruk:
Thanks, sounds a good idea to use OpenCV's conversion if it's faster, will merge now

PR #110, opened 2022-05-11
Fix some Qt complaints caused by the location of various imports

davidplowman:
Qt is really picky about importing stuff in a thread other than the
one in which you want to run its event loop. The recent import tidying
has left behind some "WARNING: QApplication was not created in the
main() thread" messages when starting the Qt preview windows.

This change moves the Qt imports out of previews/__init__.py and puts
them into previews/qt.py so that we still have a reasonably tidy way
to import them when necessary. The examples and tests, and also
previews/qt_previews.py, have been tidied to use this new import path.

### DIFF
--- a/examples/app_capture.py
+++ b/examples/app_capture.py
@@ -3,7 +3,7 @@
 from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPushButton, QVBoxLayout, QApplication, QWidget
 
-from picamera2.previews import QGlPicamera2
+from picamera2.previews.qt import QGlPicamera2
 from picamera2 import Picamera2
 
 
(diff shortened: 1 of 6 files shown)


davidplowman:
@chrisruk Just one more thing...!

Today I've noticed some Qt complaints popping out, even if you just run something like examples/preview.py. The problem is that Qt is annoyingly picky about making you import from the same thread where you run the event loop. No idea why, it seems to keep thinking it needs to protect you from yourself.

This commit is how I've tried to fix it, by not importing the Qt stuff by default in previews/__init__.py. Does this seem reasonable? I've tried to leave a reasonably tidy way to import these things (with a previews/qt.py file), and tidied up a few other files in line with this.

chrisruk:
Oops, sorry didn't spot that, yeah that looks fine to me, will merge now

PR #112, opened 2022-05-12
An extra recording example

davidplowman:
@chrisruk Are you OK with this (very) little API enhancement? In the 2nd of the two commits I added a little Qt recording app example which was in response to a forum question. I felt that always doing `picam2.encoder = encoder` seemed a bit verbose and that it might be nice to allow it to be passed in the `start_encoder` method. (Alternatively, we could always force the encoder to be passed...) Anyway let me know what you think!

### DIFF
--- a/examples/app_recording.py
+++ b/examples/app_recording.py
@@ -1,0 +1,58 @@
+#!/usr/bin/python3
+
+from PyQt5 import QtCore, QtGui, QtWidgets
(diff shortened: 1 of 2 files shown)


chrisruk:
@davidplowman Yup, I think this pull request makes a lot of sense and makes it simpler to start the encoder.
Happy for me to merge it?

PR #114, opened 2022-05-12
Changes to start_, start, and configure_.

IanTBlack:
Removed controls option from start functions and implemented set_controls in configure_. Updated relevant examples to match.

### DIFF
--- a/examples/exposure_fixed.py
+++ b/examples/exposure_fixed.py
@@ -7,9 +7,9 @@ import time
 
 picam2 = Picamera2()
 picam2.start_preview(Preview.QTGL)
-
-preview_config = picam2.preview_configuration()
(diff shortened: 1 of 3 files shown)


IanTBlack:
@davidplowman, here are the changes you requested. For the style checker, not sure why E251 was thrown. I think it only applies to function definitions.  Looks like the tests fail because generateConfiguration was changed to generate_configuration in the libcamera python bindings.

IanTBlack:
Sounds good. I need to switch gears for a few days, but next week I should be able to explore the new libcamera bindings and think about controls.

davidplowman:
I've taken this commit and patched up a couple more things to make a new PR ([here](https://github.com/raspberrypi/picamera2/pull/115)). So I'll close this one and merge the new one.

I've signed it off by both of us on this occasion, hope that's ok!

Issue #116, opened 2022-05-15
capture_jpeg.py:  numpy.ndarray size change.. Expected 48, got 40


slowrunner:
Running capture_jpeg.py throws exception:
```
$ ./capture_jpeg.py 
Traceback (most recent call last):
  File "/home/pi/Pi4BE32/Examples/picamera2/./capture_jpeg.py", line 6, in <module>
    from picamera2 import Picamera2, Preview
  File "/home/pi/picamera2/picamera2/__init__.py", line 1, in <module>
    from .picamera2 import Picamera2, Preview
  File "/home/pi/picamera2/picamera2/picamera2.py", line 14, in <module>
    from picamera2.encoders import Encoder
  File "/home/pi/picamera2/picamera2/encoders/__init__.py", line 3, in <module>
    from .jpeg_encoder import JpegEncoder
  File "/home/pi/picamera2/picamera2/encoders/jpeg_encoder.py", line 1, in <module>
    import simplejpeg
  File "/usr/local/lib/python3.9/dist-packages/simplejpeg/__init__.py", line 1, in <module>
    from ._jpeg import encode_jpeg
  File "simplejpeg/_jpeg.pyx", line 1, in init simplejpeg._jpeg
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 48 from C header, got 40 from PyObject
```

PiOS Bullseye 32-bit: Linux Pi4BE32 5.15.32-v7l+ #1538 SMP Thu Mar 31 19:39:41 BST 2022 armv7l GNU/Linux  
Python 3.9.2  
python3-libcamera/stable,now 0~git20220510+22656360-1 armhf [installed]  
v4l2-python3==0.3.1  

```
$ libcamera-jpeg --list
[0:04:16.510502904] [975]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+3544-22656360
[0:04:16.526498464] [976] ERROR CameraSensor camera_sensor.cpp:591 'ov5647 10-0036': Camera sensor does not support test pattern modes.
[0:04:16.543963437] [976]  INFO RPI raspberrypi.cpp:1356 Registered camera /base/soc/i2c0mux/i2c@1/ov5647@36 to Unicam device /dev/media3 and ISP device /dev/media0
Available cameras
-----------------
0 : ov5647 [2592x1944] (/base/soc/i2c0mux/i2c@1/ov5647@36)
[...]

PR #117, opened 2022-05-16
Update documentation and setup.py for alpha release

davidplowman:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Change Log
+
+## Unreleased
+
+## 0.2.1 Alpha Release
+
+### Added
+
+* Everything installable via apt and pip.
+* Many new examples.
+
+### Changed
+
+* Revised API - please refer to README.md file and examples.
+
+## 0.1.1 Pre-Alpha Release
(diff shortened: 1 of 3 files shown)


chrisruk:
@davidplowman Thanks, just a couple of questions:

Should numpy not be installed from apt, as it takes quite a while to compile I thought.  I guess the reason for not doing that though is due to - https://github.com/raspberrypi/picamera2/issues/116 ?

I assume we're not supporting the OS prior to bullseye then, which uses python 3.7?

Cheers

davidplowman:
> @davidplowman Thanks, just a couple of questions:
> 
> Should numpy not be installed from apt, as it takes quite a while to compile I thought. I guess the reason for not doing that though is due to - #116 ?

Yes exactly. It's been commented on in the forums as well. Maybe once we do another full Bullseye update we might be able to get rid of it.

> 
> I assume we're not supporting the OS prior to bullseye then, which uses python 3.7?

Yes, also correct. We don't update libcamera packages on Buster any more so basically everyone needs to be on Bullseye.

Thanks!

> 
> Cheers

chrisruk:
Makes sense, will merge now then :)

PR #118, opened 2022-05-17
WIP: Initial attempt to support multiple outputs

chrisruk:


chrisruk:
The idea of this change and example is to allow simultaneous streaming and motion detection/file recording.

picamera2/encoders/encoder.py
```diff
@@ -64,7 +64,11 @@ def output(self):
 
     @output.setter
     def output(self, value):
-        if not isinstance(value, Output):
+        if isinstance(value, list):
+            for out in value:
+                if not isinstance(out, Output):
+                    raise RuntimeError("Must pass Output")
+        elif not isinstance(value, Output):
```
davidplowman, line 71: Just a small thing, I wonder if this code (and all the repeats of `isintance(..., list)`) would look tidier if we started this method with
```
    if not isinstance(value, list):
        value = [value]
```
so that we can then assume we always have a list? Perhaps the "getter" could "de-listify" its return value if there's only one thing in the list.

But don't change it unless you think it actually looks more readable!

picamera2/outputs/fileoutput.py
```diff
@@ -36,7 +37,8 @@ def outputframe(self, frame, keyframe=True):
...
             self._fileoutput.flush()
 
-    def stop(self):
+    def stop(self, forcestop=False):
         """Close file handle and prevent recording"""
-        super().stop()
-        self._fileoutput.close()
+        if forcestop or not self._nostop:
+            super().stop()
+            self._fileoutput.close()
```
davidplowman, line 44: I've not quite understood what `_nostop` and `forcestop` are for. Is is so that some outputs (when you have a list of several) can be made to stop by (for example) `stop_recording` but others will carry on?

picamera2/encoders/encoder.py
chrisruk: Good idea, just changed the code to do this

picamera2/outputs/fileoutput.py
chrisruk: Yup that's right, its so others can carry on, for example the streaming file can carry on, whilst the circular buffer file output can be stopped

davidplowman: So in the example I see that the socket output is created with `nostop=True`, but I couldn't see who was ever trying to stop it. Does that mean that the parameter wasn't actually necessary?

chrisruk: Good point, I've just removed them, I think I had that in so that the encoder would stop all but ones with nostop, but we don't need that anymore because of how the example works.

davidplowman:
It's all looking good to me now. Could I trouble you for just a couple more small things:

- Could you squash the 4th and 5th commits into the first one? I think it's better simply to see the "neat final answer" than to have people looking at commits where things appear and disappear...
- And could you add a line to the CHANGELOG.md? I think this is probably something we should start doing now, especially where changes affect the external APIs or functionality. And it's probably better to do this as we go. Only I can see it's going to become a permanent rebase-conflict-headache, so not quite sure how to deal with that. Any ideas?

chrisruk:
Re. the changelog, should I add in the 0.2.1 Alpha Release section, in the changed subsection, something like:

* Support for multiple outputs for the encoder

davidplowman:
So I think the changelog has a "0.2.1 Alpha Release" section already, so I think anything new goes in the "Unreleased" section, as I did in [this PR](https://github.com/raspberrypi/picamera2/pull/120). Your one-line description looks perfect to me. When 0.2.2 gets released, we change "Unreleased" to "0.2.2" and make a new "Unreleased" on top. Though as I said, it seems to me we're permanently going to get rebase conflicts here...

chrisruk:
Gotcha, that makes sense, just updated the changelog and squashed the commits together.

I'm not sure how to avoid conflicts here unfortunately

PR #119, opened 2022-05-17
Proposed changes for libcamera configuration.

IanTBlack:
This PR contains a module and examples for modifying libcamera configuration through a variety of methods. These should be compatible with the recent changes to the libcamera python bindings and does not interfere with base picamera2 functionality. The module and examples are provided as a starting point for discussion. Please do not integrate with the main branch.

Paired with the build_libcamera_config function in the examples, the user can adjust stream and transform configuration by assigning attributes, supplying a dict, or by supplying a class with keywords. 

I believe that the addition of this configuration module would allow users to make their own configurations, as well as make current configuration self describing. I am looking for feedback and if there is interest in integrating something like this with the main branch, I am happy to coordinate with that. I believe this format would also be good for issuing controls to the camera.

Each class has annotations, which are intended to be used to check the user input against possible types.
For StreamOptions, all parameters default to None. Any parameters with a value of None are not passed to libcamera, meaning what libcamera holds as defaults will be used. It would appear the size, buffer_count, pixel_format, and color_space are required in some contexts, so I will need to build in checks for those.

I also provide example functions that acquire the camera and configure libcamera so that there is not interference with the current build. 

In the provided examples, I provide code in the build_libcamera_config function that generates stream configuration by index depending on whether or not a stream role is supplied.

If you have suggestions, please let me know,

Signed-off by: Ian Black <blackia@oregonstate.edu>


IanTBlack:
@davidplowman I've drafted  and updated some classes that should make it easier to create custom libcamera and controls configurations. The goal of this is to give users the option of inputting configuration as class attributes, as keywords, or as a dictionary. Please see the examples for how things can be configured with this new method. If you want to move forward with this, I will work on integrating with the Picamera2 class.

Stream, transform, and control options are all dynamically assigned using output from libcamera. I think this should mean that updates to libcamera streams and controls will carry over, requiring little maintenance on the python end. I did have to hardcode some things for controls and stream options, but if those are removed from libcamera they should be skipped over with no issues.

Thanks,
- Ian

IanTBlack:
I am not tied to naming conventions, either, so if there are more straightforward and descriptive names, please let me know.

davidplowman:
Hi, thanks for sending this. Mostly it looks to me like it's related to libcamera and not really to Picamera2, is that right? I'm slightly wondering whether this is the right place for it, the idea that users might be trying to configure libcamera without going through Picamera2 is a bit confusing.

As regards Picamera2, I think it would certainly be helpful to have better checking of controls and values so that it's harder make mistakes. Maybe having some kind of class representation would help? I don't really know. I'd be keen to understand and discuss what we're trying to achieve, but at the moment I don't think I've quite got it.

Thanks!

IanTBlack:
These changes are for picamera2. The additions do not configure libcamera anymore than what is currently done in the base module.  The proposed classes provide methods for creating custom configurations instead of using a prebuilt one. They would be  integrated into the Picamera2 module so that a user doesn't have to have knowledge of libcamera past that of what is already required. I think the additions that are presented could be compatible to functions like `still_configuration`, but they also leave room for custom configurations. The goal is to make camera configuration self-describing so that a user understands what is being changed where.

So if someone where to make a custom configuration, they could do something like this (of the three possible methods for configuring a camera).

```
config = CameraOptions()

config.main.size = (1280,720) #I am configuring the main stream size to be (1280,720)
config.transform.hflip = 1   # I am configuring the image orientation(transform) by flipping on the horizontal axis.
config.controls.Brightness = 1  # I am configuring the camera brightness control to 1.
```

Then `config` would be passed to `picam2.configure()`.

I attempted to make them dynamic because you previously expressed concern about future maintenance and new control additions. I think that it would be better to hardcode things like the control names so that it is easier for end users, particularly python users with no C++ background. Of course the downside to this is that patches to picamera2 will need to be made along with libcamera, but that would  probably need to happen anyway if picamera2 is going to use newer versions of libcamera.

I would be happy to work on type checking for stream and control configuration, but I would need confirmation from you that it is acceptable to hardcode things like `buffer_count` and `brightness`.

I think a possible alternative to all of this would be to setup libcamera yaml files for stream and control configuration (like the one you showed previously). These could be setup to denote valid types, ranges, etc. Then Python could read the yaml files in when Picamera2() is instantiated and determine valid stream and control configs from there.

davidplowman:
Thanks for explaining. I might feel more comfortable if we can avoid examples that are configuring libcamera directly and "going behind Picamera2's back", I would worry that users would start thinking this was what they're "meant" to do!

I'd also like to understand what the typical use patterns might be. For example, for a user who just wants to throw up a preview window, what would they do? Where would the defaults come from? Currently that's done by having some notion of a "use-case", like preview, still or video. Perhas `CameraOptions(Role.PREVIEW)` or something like that. For the still one, the default resolution comes from the camera itself, so we'd have to think how that would work. (I'm also not wild about the word "options" as it tends to get over-used, but not sure what is better!)

Another thing to be understand is how users would say whether they want extra low resolution or raw streams. Maybe some examples up-front of the API before implementing too much stuff would be helpful here.

One final point is that not all the configuration items are necessary meaningful to libcamera. For example the "display" and "encode" fields are meaningful to Picamera2 but not to libcamera. So again, this probably comes back a bit to the question of whether we're making something that would make sense to libcamera as well as Picamera2. Lots to think about!!

PR #120, opened 2022-05-18
More generalised callbacks within the camera event loop

davidplowman:
This work was prompted by some forum question where I thought "Picamera2 would be a good solution for that", only it wasn't quite!

Specifically, this lets you do things like putting date/timestamps onto the frames that get recorded by the video encoder, although it is a bit more flexible than that. There are two commits:

* The first allows applications to define 3 callback functions, one of which is what the old "request_callback" mechanism used to do.
* The second commit adds a couple of examples, including the timestamps-in-a-video-recording one.


davidplowman:
Changed my mind a bit. Only 2 callbacks make sense. There's no point having another one after sending stuff to the video encoder because then you've got a race condition on that image buffer. And if you're not running an encoder, then the two things are equivalent anyway.

chrisruk:
Looks good to me! I think having just 2 callbacks sounds a good idea, and the timestamp example looks cool.

Originally I thought you meant timestamping frames for 'seeking'.

Will merge now.

PR #122, opened 2022-05-19
Update issue templates

davidplowman:
I've created some issue templates in order to make reporting and tracking easier. We have:

Bug reports
Suggestions or feature requests
How-To questions
And all other types of issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,28 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: ''
+assignees: ''
+
+---
+
+**Please only report one bug per issue!**
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behaviour. Wherever possible, please include a small self-contained Python script that shows the bug.
+
+**Expected behaviour**
+A clear and concise description of what you expected to happen.
+
+**Console Output, Screenshots**
+If applicable, any console output or screenshots that show the problem and associated error messages.
+
+**Hardware :**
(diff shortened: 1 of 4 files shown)


tvoverbeek:
What about a separate category for documentation?
Case in point: The README for the Tensorflow examples contains the old build instructions, should be updated to follow the main README.

davidplowman:
Yes, documentation is an interesting question.

So the plan is that the documentation here in the repository is basically going to vanish. Maybe just a few things will remain, a link to the new documentation, the contribution guidelines and so on.

Then there'll be a proper "Picamera2 Manual" written from scratch that will be much more extensive than what's here now. Once that exists we'll make sure we have some way to report errors in it, maybe that will be issues here. Do you think it's worth having a documentation category in the interim for the old readme? I'm not sure. I suppose it would stop people wondering why there's no documentation category...!

I note that you're right about that tensorflow readme, though. It's quite out-of-date and actually wrong in many respects. It probably just wants deleting because everything is covered in the main readme.

Thanks!

tvoverbeek:
> Do you think it's worth having a documentation category in the interim for the old readme? I'm not sure.

Agreed. My suggestion was triggered by the outdated Tensorflow README.
Will the official documentation have its own repository or be part of the overall documentation repository?

I guess in the interim we could report those type of issues in the 'Other' category.

davidplowman:
> My suggestion was triggered by the outdated Tensorflow README. 

There's now a PR to get rid of that, so thanks for pointing it out!

>Will the official documentation have its own repository or be part of the overall documentation repository?

That's up to the documentation overlords, of whom I am not one... hopefully I'll know more soon.

chrisruk:
@davidplowman They all look good to me, I think keeping an issue to a single item makes a lot of sense.  Happy for me to merge now?

PR #126, opened 2022-05-20
WIP: Make GUI dependencies optional

chrisruk:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,7 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-from setuptools import setup, Extension
-from os import getenv
+from setuptools import setup
 
 with open("README.md") as readme:
     long_description = readme.read()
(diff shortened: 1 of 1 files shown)


chrisruk:
The problem I can see with this approach is that then users need to specify they want the GUI deps via pip

davidplowman:
That's so nearly exactly what we need. Except we want "GUI" to be "on" by default, or a "NOGUI" option instead that can "remove" dependencies. How annoying!

chrisruk:
Yeah it's a pain heh, I also looked at https://stackoverflow.com/questions/18725137/how-to-obtain-arguments-passed-to-setup-py-from-pip-with-install-option but noticed this warning

```
pip3 install . --user --install-option="--nogui"
/usr/lib/python3/dist-packages/pip/_internal/commands/install.py:245: UserWarning: Disabling all use of wheels due to the use of --build-option / --global-option / --install-option.
```

So assume it wouldn't work with a pypi wheel.    

Going to push another pull request with an alternative idea.

PR #127, opened 2022-05-20
Update camera_properties after configure

davidplowman:
Some camera properties can change after a call to configure, as they
depend on the selected camera mode.


chrisruk:
Just wondering if you should create a @property for camera_properties, that proxies camera.properties?

davidplowman:
Ah yes, some of this Python stuff is still a bit alien to me! Let me try that and re-post this...

PR #128, opened 2022-05-20
Make GUI dependencies optional

chrisruk:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,13 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-from setuptools import setup, Extension
-from os import getenv
+import os
+from setuptools import setup
+
+reqs = ['numpy', 'PiDNG', 'piexif', 'pillow', 'simplejpeg', 'v4l2-python3', 'python-prctl']
+allreqs = reqs + ['pyopengl', 'PyQt5']
+if os.getenv('NOGUI', '0') == '1':
+    allreqs = reqs
 
 with open("README.md") as readme:
     long_description = readme.read()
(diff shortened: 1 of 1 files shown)


chrisruk:
Fingers crossed that this works once uploaded to pypi, will need to test that

davidplowman:
Ah cool, I guess I can test this with a clean image and trying `pip3 install git+https://github.com/RaspberryPiFoundation/picamera2.git@optional2` and then repeat with a Lite image and the `NOGUI=1 ...` version.

davidplowman:
I've just tried all that and it all worked great, with the only minor exception being that on Raspberry Pi OS Lite I had additionally to run `sudo apt install -y libopenjp2-7`. But then it was fine.

Could I trouble you to add a 2nd commit that updates the README to explain how to do a "NOGUI" install, and add a quick line to the CHANGELOG - I think it's a sufficiently obvious/beneficial change that users would want to know!

tvoverbeek:
I assume eventually these will end up as debian packages.
Why don't you split it in 2 packages:
picamera2-base which contains the non-gui stuff (and maybe the non-gui examples) and a picamera2-gui which contains
the stuff needed for the GUI (and examples) and which requires picamera2-base.
Installing picamera2-gui is then equivalent to the current setup.

chrisruk:
I think we do need to also provide a way for people to choose whether to install GUI deps or not via pypi though.

@davidplowman Are you ok if I remove the use of 'sudo' for pip3 installs, as I think generally using sudo for pip might be slightly frowned upon e.g - https://askubuntu.com/questions/802544/is-sudo-pip-install-still-a-broken-practice

davidplowman:
Yes, sounds good to me.

I agree that being able to choose the GUI or not via pip is necessary. Default should be to install the GUI stuff, I think, which is how you have it already.

I'll have a word with my person who deals with apt for me about this. For our libcamera-apps packages, for example, we make two packages from the one repo, a "libcamera-apps" and "libcamera-apps-lite" which omits all the GUI stuff in the same way.

davidplowman:
@chrisruk Are we ready to merge this one?

PR #133, opened 2022-05-23
WIP: Initial attempt to fix deque mutated with circular buffer

chrisruk:


chrisruk:
@davidplowman Good point, I'll squash them together, need to do a bit more testing on this though

chrisruk:
@davidplowman I've just squashed the commits together and haven't had the 'deque mutated' issue, so think it's good to go

PR #136, opened 2022-05-24
Properly sync audio

dinapappor:
This properly syncs up audio, by using `-use_wallclock_as_timestamps 1` for both audio _and_ stdin video input streams. This makes it easier for ffmpeg to sync audio. And I _think_ we can remove `-itsoffset` here also, at least it was possible in a project of [mine](https://github.com/autosbc/rpidashcam/blob/master/dashcam.py#L284) (based on mmalobj.py)

### DIFF
--- a/picamera2/outputs/ffmpegoutput.py
+++ b/picamera2/outputs/ffmpegoutput.py
@@ -60,6 +60,7 @@ class FfmpegOutput(Output):
                            '-f', 'pulse',
                            '-sample_rate', str(self.audio_bitrate),
                            '-thread_queue_size', '512',  # necessary to prevent warnings
+                           '-use_wallclock_as_timestamps', '1',
                            '-i', self.audio_device]
             audio_codec = ['-b:a', str(self.audio_bitrate),
                            '-c:a', self.audio_codec]


dinapappor:
Also, I noticed that unless I specify my usb microphone, which can be found with `arecord -l` (and then specified with `hw:2,0`) I would get a `no such process`/`process not found` error. Should I add a note about that in `examples/audio_video_capture.py`  as well?

davidplowman:
Hi, thanks for sending this. I gave it a try but I get lots of warnings while it runs:
```
[aac @ 0xcdff60] Queue input is backward in time
[mp4 @ 0xc8a2b0] Non-monotonous DTS in output stream 0:1; previous: 88511, current: 87862; changing to 88512. This may result in incorrect timestamps in the output file.
[aac @ 0xcdff60] Queue input is backward in time
[mp4 @ 0xc8a2b0] Non-monotonous DTS in output stream 0:1; previous: 88512, current: 87951; changing to 88513. This may result in incorrect timestamps in the output file.
... (forever)
```
Did you get these too?

dinapappor:
@davidplowman No I did not. :thinking: 

I am using 
```
ffmpeg version 4.3.4-0+deb11u1+rpt1 Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 10 (Debian 10.2.1-6)
  ```
  is picamera2/libcamera by chance adding timestamps as well?

davidplowman:
Hmm, not sure what's happening there. I did pull your code down to be sure I was running the same thing, but I get the same errors. My FFmpeg version is identical to yours. Perhaps I shall try it on a fresh Bullseye installation when I get the chance.

As regards timestamps, FFmpeg generates them all for itself. For the audio... it does whatever it normally does. For the video, while picamera2 knows accurate timestamps I don't have any easy way of passing those to FFmpeg, so I have to get it to re-timestamp them when it receives them. There will be some resulting jitter in the video frame times because of this.

davidplowman:
I have just tried this on a completely fresh installation, again with the same results. The output file seems OK, it's just all those warnings that are rather alarming!

davidplowman:
Ah, I see, so that's the difference then (I won't ask why pulse audio doesn't work for you...!)

Would you be able to convert the PR so that it leaves pulse alone but adds the parameters you want (or at least, gives you a way to add them) for your case? I think that might then work for everyone.

dinapappor:
I'll try to get pulse working and see if I can make it work for both cases. :)

davidplowman:
Hi, I think I'm probably going to close this one for now. But if you do have anything you'd like us to look at down the line, please file another PR or create an issue where we can discuss things. Thanks!

3drs:
Ref OP "Also, I noticed that unless I specify my usb microphone, which can be found with arecord -l (and then specified with -i hw:2,0) I would get a no such process/process not found error. Should I add a note about that in examples/audio_video_capture.py as well?"

How do we specify hardware/mic in picamera2? So the "libcamera -i hw:2,0" equivalent in picamera2.

I can't find it in the manual (https://datasheets.raspberrypi.com/camera/picamera2-manual.pdf)

PR #143, opened 2022-05-27
Preview rework

davidplowman:
These commits involve quite a lot of rework to the various preview window implementations. So there are quite a lot of commits here, my apologies for that. There are

- 4 commits related to the QGlPicamera2 widget, the last of which just amends a couple of tests to work better;
- 3 commits for the DrmPreview;
- 1 commit for the QPicamera2 and 1 more to add another example;
- And then 1 final commit that simply updates the changelog.

In all cases, the commits make the previews handle aspect ratio correctly. Image aspect ratios are now preserved by default (though that can be over-ridden). Also the treatment of overlays is improved in all cases so that they get redrawn even if the camera is not delivering images. Finally, the Q versions also allow the background colour to be set (as this may be visible if the widgets and camera images do not share the same aspect ratio).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Changed
 
+* The preview window implementations are changed to preserve image aspect ratios which they did not previously. The Qt variants also respect resize events.
+
 ## 0.2.1 Alpha Release
 
 ### Added
--- a/examples/app_capture.py
+++ b/examples/app_capture.py
@@ -26,7 +26,7 @@ def on_button_clicked():
         print("Busy!")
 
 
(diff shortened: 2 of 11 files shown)


jlprojects:
Hi,
Yes - absolutely fine. I've not had the chance to any more work on it recently. Just checked out this branch, and it seems to be working fine for me. 
I'm perfectly happy for you to use my contribution. Glad it's been of some use. :)

davidplowman:
Well, I guess so. It's really all way too much of a change to comb through carefully and understand, so I think we're just going to have to merge it and see if there's any fall out.

PR #149, opened 2022-05-30
WIP: UDP Example

chrisruk:
Sends UDP packets of camera stream to remote host.

Can play with:

```
mpv udp://127.0.0.1:10001
```

Will work on catching exceptions when connection terminates.

Also need to work out how to play UDP from server

### DIFF
--- a/examples/capture_stream_udp.py
+++ b/examples/capture_stream_udp.py
@@ -1,0 +1,25 @@
+#!/usr/bin/python3
+
+import socket
+import time
+
+from picamera2.encoders import H264Encoder
+from picamera2.outputs import FileOutput
+from picamera2 import Picamera2
+
+picam2 = Picamera2()
+video_config = picam2.video_configuration({"size": (1280, 720)})
+picam2.configure(video_config)
+encoder = H264Encoder(1000000)
+
+with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
+    sock.connect(("REMOTEIP", 10001))
+    picam2.encoder = encoder
+    stream = sock.makefile("wb")
+    picam2.encoder.output = FileOutput(stream)
+    picam2.start_encoder()
+    picam2.start()
+    time.sleep(20)
+    picam2.stop()
+    picam2.stop_encoder()
+    conn.close()
--- a/examples/capture_stream_udp.py
+++ b/examples/capture_stream_udp.py
@@ -14,12 +14,7 @@ encoder = H264Encoder(1000000)
 
 with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
     sock.connect(("REMOTEIP", 10001))
-    picam2.encoder = encoder
     stream = sock.makefile("wb")
-    picam2.encoder.output = FileOutput(stream)
-    picam2.start_encoder()
-    picam2.start()
+    picam2.start_recording(encoder, FileOutput(stream))
     time.sleep(20)
-    picam2.stop()
-    picam2.stop_encoder()
-    conn.close()
+    picam2.stop_recording()
--- a/examples/capture_circular_stream.py
+++ b/examples/capture_circular_stream.py
@@ -34,14 +34,14 @@ def server():
         sock.bind(("0.0.0.0", 10001))
         sock.listen()
         while tup := sock.accept():
+            event = threading.Event()
             conn, addr = tup
             stream = conn.makefile("wb")
             filestream = FileOutput(stream)
             filestream.start()
             picam2.encoder.output = [circ, filestream]
-            while not filestream.dead:
-                time.sleep(1)
-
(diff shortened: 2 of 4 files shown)


examples/capture_stream_udp.py
```diff
@@ -0,0 +1,25 @@
...
+picam2.configure(video_config)
+encoder = H264Encoder(1000000)
+
+with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
+    sock.connect(("REMOTEIP", 10001))
+    picam2.encoder = encoder
+    stream = sock.makefile("wb")
+    picam2.encoder.output = FileOutput(stream)
+    picam2.start_encoder()
+    picam2.start()
```
davidplowman, line 21: I think these 3 lines above could just be replaced by `picam2.start_recording(encoder, output)`. Similarly for the stopping code just below.

davidplowman:
BTW, I didn't think we'd get to know if the client stops receiving, the point of UDP being that it's "fire and forget" (and no guarantee of arrival). Or is that not the case here?

chrisruk:
> BTW, I didn't think we'd get to know if the client stops receiving, the point of UDP being that it's "fire and forget" (and no guarantee of arrival). Or is that not the case here?

Good question, when I close mpv before the stream ends, I get this thrown by picamera:

```
Exception in thread Thread-1:
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/pi/Repositories/picamera2/picamera2/encoders/v4l2_encoder.py", line 173, in thread_poll
    self.outputframe(b, keyframe)
  File "/home/pi/Repositories/picamera2/picamera2/encoders/encoder.py", line 97, in outputframe
    out.outputframe(frame, keyframe)
  File "/home/pi/Repositories/picamera2/picamera2/outputs/fileoutput.py", line 37, in outputframe
    self._write(frame)
  File "/home/pi/Repositories/picamera2/picamera2/outputs/fileoutput.py", line 47, in _write
    self._fileoutput.flush()
  File "/usr/lib/python3.9/socket.py", line 722, in write
    return self._sock.send(b)
ConnectionRefusedError: [Errno 111] Connection refused
```

Need to think about how to deal with this

davidplowman:
I've been looking into all this a bit. For a UDP socket, the `connect` method merely sets a "default" destination, so that you can use functions like `send` and `write`, though you can still send packets elsewhere using `sendto` if you want. That all seems reasonable enough.

Now, the "connection refused" messages are being signalled back by ICMP and generally I think folks tend to ignore them. The behaviour is then that you can start your UDP server, and on the client end there might be no one listening, but that shouldn't really matter, it just sends anyway. One thing to bear in mind is that you might be sending to a multicast address, so you can't really have any strong expectations as to exactly how many clients there might be (including none).

So my impression is that we should really be ignoring these "connection refused" errors entirely, and we carry on transmitting on the grounds that someone or other may start receiving at a later point. Possibly we should ignore these errors only if the underlying transport is a UDP socket, I'm not clear if this error could come about from other sources that might be "genuine" errors.

What do you thinik?

chrisruk:
Yeah I agree that makes sense about ignoring the error.  I do set a dead flag that can be checked if the user wants to as well.

I think I have to catch the error for TCP sockets too, as there would be no way for the user to catch that themself as it happens in its own thread?

davidplowman:
Yes, setting a "dead" flag and otherwise ignoring the error seems fine for UDP. The application can either check it or ignore it. I suppose for TCP it's OK too, though I don't really see any benefit it continuing to try and send stuff. That makes sense for UDP but for TCP it can presumably never recover. Instead, an application would have to go back to a state where it waits for a new connection?

chrisruk:
https://github.com/raspberrypi/picamera2/blob/main/examples/capture_circular_stream.py#L30 has an example of how I made it wait for another connection with TCP

I am just thinking if the FileOutput could in future maybe have a callback mechanism to inform the user, when a connection dies, rather than them having to check a flag.

davidplowman:
Looking pretty good to me now. At the risk of being super-picky, I wonder about including "_callback" in the field name for the callback function that we can now set? It makes it a bit more obvious you need to put a function or lambda in there. Also, is there any sense passing the error object into that callback function, just in case a caller might care what kind of error it was?

chrisruk:
Good points, just renamed the argument to _callback, and now the exception will be passed as an argument to the callback

davidplowman:
Hi Chris, is this in a state where you're happy to merge it?

chrisruk:
@davidplowman Yup, I've just made one small extra change, as I was getting a brokenpipeerror, when attempting to close the stream, if it was already closed, but I've caught that now.

So it's ok from my perspective now, let me know if there's any changes you'd like making though.

PR #150, opened 2022-05-30
Update README.md

ibmua:
The obvious. pip3 doesn't exist till it's installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ All the necessary packages can now be installed via `apt` and `pip3`, so the fol
 
 ```
 sudo apt install -y python3-libcamera python3-kms++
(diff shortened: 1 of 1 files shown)


davidplowman:
Thanks for pointing this out. I'm guessing this is only a problem with Raspberry Pi OS Lite? But it seems harmless to list it even when it's already installed, so I'm good to merge this. Thanks!

davidplowman:
Just as a note, we normally ask for contributions to be "signed" according to the [guidelines](https://github.com/raspberrypi/picamera2#contributing), but on this case I think it's sufficiently trivial that we can let it pass!

ibmua:
Yeah, this was based on Pi Lite experience.

PR #161, opened 2022-06-07
Add YUV420 support to the QPicamera2 widget

davidplowman:
This is the non-OpenGL accelerated version which previously could not
display YUV420 format images. It can do so now through the use of
OpenCV which provides a much faster YUV to RGB conversion function
than we can write easily using numpy. (But be under no illusions that
it still costs quite a bit of CPU!)

If OpenCV is not available, an error is only thrown if you try to
display a YUV420 image, everything else will work as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+* The QPicamera2 widget will now display YUV420 images if the cv2 (OpenCV) module is installed.
 * Methods that allow still captures to be saved to files have been updated to accept file-like objects, such as BytesIO.
 * NOGUI=1 environmental variable to install without GUI dependencies
 * The Picamera2's request_callback has been changed to post_callback. A new pre_callback has been added which runs before images copied for applications.
--- a/picamera2/previews/q_picamera2.py
+++ b/picamera2/previews/q_picamera2.py
@@ -5,6 +5,13 @@ from PyQt5.QtGui import QBrush, QColor, QImage, QPixmap, QTransform
 import numpy as np
 
 
+try:
+    import cv2
+    cv2_available = True
+except ImportError:
+    cv2_available = False
+
+
 class QPicamera2(QGraphicsView):
     done_signal = pyqtSignal()
     update_overlay_signal = pyqtSignal(object)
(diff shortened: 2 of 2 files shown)


chrisruk:
@davidplowman Was just wondering should it fallback to your numpy yuv420 -> rgb converter if opencv is not found?
Or is that too slow for this?

davidplowman:
That's a fair question. My testing showing that the numpy version is excruciatingly slow compared to the OpenCV one, and I'm talking a factor of 10. I don't really understand why the numpy one is so bad - I don't think it's doing anything obviously stupid - or why the OpenCV one is so much better. If anyone out there understands, I'd love to know why!

Anyway, my gut instinct is to avoid a fallback behaviour that is (IMHO) so bad that folks will just complain about it. On the other hand there'll be Pi Zero users out there who don't want to install OpenCV at all. But then if they want any kind of acceptable preview performance at all, then they really should run without X-Windows and use the DRM preview (which supports YUV420 anyway).

So lots of swings and roundabouts. But for the moment I'm inclined to think that failing if OpenCV isn't there seems reasonable...

chrisruk:
Cheers, thanks for explaining that, will merge now

BrodyStone21:
Do you have any suggestions for YUV420 support with the [picamera2/examples/mjpeg_server.py](https://github.com/raspberrypi/picamera2/blob/main/examples/mjpeg_server.py) script?

davidplowman:
> Do you have any suggestions for YUV420 support with the [picamera2/examples/mjpeg_server.py](https://github.com/raspberrypi/picamera2/blob/main/examples/mjpeg_server.py) script?
> 
> I've been trying to get it working for quite a while and I'm not sure what I'm getting wrong, but I can not get it to work. I'm using my Raspberry Pi 4 as a camera server, which I'm connecting to from my Ubuntu server, but my Nvidia GPU is unable to decode yuv444.

Hi, I'm not quite sure what you mean by "YUV420 support" here. The script you reference obviously works as far as it goes, because I can point Chrome or Firefox at it and it seems to play the stream correctly. The encoded JPEG images are inherently "YUV420", even though Python encoders normally want RGB which they then have to convert. I don't know anything about Nvidia GPUs and so can't really help you with that. Maybe capture one of these encoded JPEGs that it doesn't handle and ask on an Nvidia forum? Sorry not to be more help.

PR #165, opened 2022-06-09
WIP: Linting

chrisruk:


chrisruk:
Good question, in the flake8 config file I set the max line length to 137.  (I'm open to changing that to higher if you want)

There seems to only be 7 errors for this:

```
7     E501 line too long (132 > 127 characters)
```

davidplowman:
That seems pretty reasonable to me!

chrisruk:
@davidplowman Hey, are you happy for me to continue adding the kind of doctrings I've added in the last commit in this pull request.  I'm using vscode + autodocstring, which creates a placeholder I think fill in.  (Using the sphinx docstring format)

davidplowman:
Yes, I think they're looking good, actually!

davidplowman:
@chrisuk Basically I just want to merge this lot because we need to move forward with all this, but I'd quite like the checks to be "clean". As regards the style checker complaints, are those fixable, or is it fundamentally disagreeing with flake8? Also all those flake8 build test errors - anything we can do about those?

chrisruk:
@davidplowman Some of the style checker errors seem to be in:
picamera2/previews/q_gl_picamera2.py , is that due to 'os.environ["PYOPENGL_PLATFORM"] = "egl"' being above the imports maybe.

A lot of the flake8 errors seem to be due to missing docstrings, which might take a while to fix.

davidplowman:
> @davidplowman Some of the style checker errors seem to be in: picamera2/previews/q_gl_picamera2.py , is that due to 'os.environ["PYOPENGL_PLATFORM"] = "egl"' being above the imports maybe, which I guess it has to be?

Yes, I think it does. I guess one benefit of the style checker is that is runs on the changes, so we can ignore complaints like that and they won't keep coming back forever (unless we change stuff around that spot again).

> 
> And I think I broke some of the lines, so that it didn't go over the max line length

Can we patch those up with some extra brackets, or by adding a `\`?

> 
> A lot of the flake8 errors seem to be due to missing docstrings, which might take a while to fix.

Maybe we should, with this PR at least, avoid running these failing checks automatically until we've had time to fix all that. But I'd still qutie like to merge this (whilst avoiding the failing checks) just so we feel we're getting somewhere!! Does that sound doable?

chrisruk:
That sounds a good plan about making flake8 exit cleanly at the moment, I've just made it still run, but exit a 0 code, so we get a nice tick next to it :)

I'm not sure how I edit previous commits though with git, to fix the style checker issues though afraid.

chrisruk:
Sure, squashing definitely makes sense! 

How about for the commit message:

Github action for flake8 linting along with initial linting of code, including addition of docstrings and run of isort on the python files.  Compilation of libcamera is also performed to allow flake8 to import it.

chrisruk:
@davidplowman I'm rather stuck with the style check at the moment I've tried a number of different ways
to break the line, all of which it doesn't seem to like alas:

```
------------------------------------------------------------------------------------------------------------
367b6cf221bd26ecf53021db364391104586018e Add configuration for pylint and sphinx docstrings for picamera2.py
------------------------------------------------------------------------------------------------------------
--- picamera2/picamera2.py
+++ picamera2/picamera2.py
#152: : E502 the backslash is redundant between brackets
+        return (self._preview is not None and getattr(self._preview, "thread", None) is not None \
---
1 potential issue detected, please review
```

```
        return self._preview is not None and getattr(self._preview, "thread", None) is not None \
        and self._preview.thread.is_alive()
```
```
------------------------------------------------------------------------------------------------------------
f917ed2194d95a8d283b461596dd3cdf68f2fb2c Add configuration for pylint and sphinx docstrings for picamera2.py
------------------------------------------------------------------------------------------------------------
--- picamera2/picamera2.py
+++ picamera2/picamera2.py
#153: : E122 continuation line missing indentation or outdented
+        and self._preview.thread.is_alive()
---
1 potential issue detected, please review
```

davidplowman:
Well, let's just go with it as it is then, we can look again at those if we ever edit the offending lines again!

tvoverbeek:
Note: line breaks between parentheses are allowed in python.
So for example try this:
```
return self._preview is not None and getattr(self._preview, "thread", 
           None) is not None and self._preview.thread.is_alive()
```

davidplowman:
> Note: line breaks between parentheses are allowed in python. So for example try this:
> 
> ```
> return self._preview is not None and getattr(self._preview, "thread", 
>            None) is not None and self._preview.thread.is_alive()
> ```

Hehe. Amusingly that does work, though it wants me to align the `None` under the `self` just after `getattr(`, so it still looks a bit weird. Actually it seems happy with this too:
```
    return self._preview is not None and \
        getattr(self._preview, "thread", None) is not None and \
        self._preview.thread.is_alive()
```
so I'll probably go with that!!

davidplowman:
> @davidplowman I'm not sure why the smoke tests failed this time though? I'm sure they where working earlier.

Looks like a classic case of the known problem. A stale request from the previous configuration ends up "lurking" inside libcamera, and pops out after we restart the camera, expecting the new resolution - so it all falls over. It's just an intermittent timing related thing, unfortunately.

I'm a bit reluctant to put in a bodge which we'll only want to undo next week when we get a proper fix from the libcamera team, but I appreciate that it is getting a bit annoying... :(

davidplowman:
> Ah cheers. Shall I amend this commit to use the following then

No, don't worry about that. I'm squashing everything together, fixing those last things up and rebasing all my own PRs to check my world is still in one piece! Once I've done that I'll push all this directly onto "next" and close this PR. Thanks!!

davidplowman:
I've pushed this up directly now, so closing. Please shout if everything is horribly broken!

PR #170, opened 2022-06-14
Update picamera2 with latest API changes from libcamera.

naushir:
Currently working at libcamera commit b4d4b78c82a23de3c6dd1676bf43bba1d6043652

Signed-off-by: Naushir Patuck <naush@raspberrypi.com>

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -86,7 +86,7 @@ class Picamera2:
     def _reset_flags(self) -> None:
         self.camera = None
         self.is_open = False
-        self.camera_controls = None
+        self.camera_ctrl_info = {}
         self._preview = None
         self.camera_config = None
         self.libcamera_config = None
@@ -110,6 +110,7 @@ class Picamera2:
         self.completed_requests = []
         self.lock = threading.Lock()  # protects the functions and completed_requests fields
         self.have_event_loop = False
+        self.camera_properties_ = {}
 
     @property
     def request_callback(self):
@@ -128,7 +129,7 @@ class Picamera2:
 
     @property
     def camera_properties(self) -> dict:
-        return {} if self.camera is None else self.camera.properties
+        return {} if self.camera is None else self.camera_properties_
 
     def __enter__(self):
         return self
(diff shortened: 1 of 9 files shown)


picamera2/picamera2.py
```diff
@@ -128,7 +129,7 @@ def asynchronous(self) -> bool:
 
     @property
     def camera_properties(self) -> dict:
-        return {} if self.camera is None else self.camera.properties
+        return {} if self.camera is None else self.properties
```
davidplowman, line 132: I think it would be more usual to use `self.camera_properties_` rather than `self.properties`.

picamera2/picamera2.py
```diff
@@ -268,8 +276,8 @@ def preview_configuration(self, main={}, lores=None, raw=None, transform=libcame
         lores = self.make_initial_stream_config({"format": "YUV420", "size": main["size"]}, lores)
         raw = self.make_initial_stream_config({"format": self.sensor_format, "size": main["size"]}, raw)
         # Let the framerate vary from 12fps to as fast as possible.
-        controls = {"NoiseReductionMode": libcamera.NoiseReductionMode.Minimal,
-                    "FrameDurationLimits": (self.camera_controls["FrameDurationLimits"][0], 83333)} | controls
+        controls = {"NoiseReductionMode": libcamera.controls.draft.NoiseReductionModeEnum.Minimal,
+                    "FrameDurationLimits": (self.camera_ctrl_info["FrameDurationLimits"][1].min, 83333)} | controls
```
davidplowman, line 280: I'd like to do something for our users about this nasty long `libcamera.controls.draft` thing, but that's probably for a subsequent commit!

picamera2/picamera2.py
```diff
@@ -128,7 +129,7 @@ def asynchronous(self) -> bool:
 
     @property
     def camera_properties(self) -> dict:
-        return {} if self.camera is None else self.camera.properties
+        return {} if self.camera is None else self.camera_properties
```
davidplowman, line 132: Does this work? I was expecting it to return `self.camera_properties_` rather than `self.camera_properties`, which feels like it might be an infinite recursion once `self.camera` is not `None`.

chrisruk:
To get the actions to run for this pull request, I think you'll need to add the branch 'next' to both
the config files in - .github/workflows

davidplowman:
Closing in order to force tests to run.

PR #171, opened 2022-06-14
Proposed refactor of PiCamera2 API.

gkyle:
With these changes, various methods for retrieving or saving a buffer or image
are replaced by switch_mode_and_capture_buffer(). This method now returns one
or more buffers + metadata. Resulting buffers can then be converted to images
or saved via methods on the PostProcess class.

I am proposing this change because I believe a typical use case for capturing
stills will perform multiple operations on the available streams (eg. convert
RGB data for display, then save a jpeg and/or DNG file). The current API only
suppoorts this through capture_request(), which requires managing the lifecycle
of CompletedRequest, so this alternative should be simpler and may offer
additional flexibility.

Existing API:
```
    camera.switch_mode(capture_config)
    request = camera.capture_request()
    pilImage = request.make_image("main")
    # optionally, display this image to UI
    request.save("main", "full.jpg")
    request.save_dng("full.dng")
    request.release()
    camera.switch_mode(preview_config)
```

Alternative API:
```
    buffers, metadata = camera.switch_mode_and_capture_buffer(capture_config, ["main", "raw"])
    pilImage = converter.make_image(buffer, capture_config["main"])
    # optionally, display this image to UI
    # optionally, defer these operations without regard for the lifecycle of CompletedRequest
    postProcessor.save(pilImage, metadata, "out.jpg")
    postProcessor.save_dng(buffers[1], metadata, capture_config["raw"], "out.dng")
```


davidplowman:
Hi, thanks for sending this, it seems like a good idea. I wonder whether it might be worth adding new functions to implement this behaviour, that way we don't break existing code and you don't have to fix all the examples!! :)  Maybe we can just make the names of the things we're capturing plural?
```
capture_buffers
capture_arrays
switch_mode_and_capture_buffers
switch_mode_and_capture_arrays
```

(I think "array" as well as "buffer" versions might be nice?)

Maybe we could even then implement the existing functions in terms of the new ones as they're more general...

On a housekeeping note, folks pointed out that we were being a bit unkind by doing development on the main branch and then breaking all the examples there for those who were using the released version. So new development has moved to the "next" branch, and we'll push that across when we do a new release. Unfortunately the "next" branch has brought us up-to-date with the mainline libcamera Python bindings which have changed quite a bit - so you would need to check out and build your own version of libcamera. You can obtain our "stable" version of their latest code [here](https://github.com/raspberrypi/libcamera) (on the master branch, the "picamera2" branch contains the older version that supports our currently released Picamera2).

Also be sure to check you've read the [contribution guidelines](https://github.com/raspberrypi/picamera2#contributing) to make sure you're happy with them. I don't think there's anything too controversial. This is not a massive sprawling commit, but a nice commit message which has been "signed" would be helpful. Also, maybe add a second commit that adds at least one example of how to use the APIs? We should add that to the list of automatic tests too (tests/test_list.txt). Documentation is being completely rewritten (when I have time!), so you don't need to do anything about that, I will get round to it.

Sorry if there are a few fiddly little obstacles to deal with, but I do like the feature and would like to see it progress. Thanks again!

davidplowman:
Thanks for the update.

I like that we now have these things in new functions, but I apologise if I wasn't clear enough, part of the reason was so that we don't have to change existing code or examples. Instead I was thinking that we could add some new examples.

One other thing I'm not sure about is the "post processor". It sounds a bit like the post-processor in our libcamera-apps and of course it isn't related at all, and I'm a little worried about people thinking "what's this post-processor thing that I have to learn about now?"

I was wondering whether it might look less intimidating if we embedded the post-processor into the Picamera2 object. We might even call it "helpers", so that folks could write things like
```
picam2.helpers.save_dng(...stuff...)
```
It would also be easy to update the request.save_dng (for example) to call this "helper" version. What do you think? Sorry if this is turning into a longer saga than you imagined...

gkyle:
Okay, no problem. I misunderstood your earlier comment about preserving these existing interfaces.

davidplowman:
Hi again, sorry for the long delay (lots of other stuff, summer holidays...).

I've taken the code and brought it all up to date, there's quite a lot happened in the meantime so loads of conflicts and stuff. I've also reorganised the changes into 3 commits that I think give a reasonably understandable history of what's happened.

As you've done lots of this work I'd like to put you as the author of these 3 commits, and add your "Signed-off-by" tag if you're happy to do that (according to the [contribution guidelines](https://github.com/raspberrypi/picamera2#contributing)). Can you let me know your email address so that I can do that? Then I'll post an updated PR for you to look at. Thanks!

davidplowman:
@gkyle Hi, sorry to ping you on this, I just wanted to check if you'd had a chance to read my latest update. Thanks!

gkyle:
Hi David. Sounds good. E-mail is kyle.scholz@gmail.com.

davidplowman:
I'm going to close this PR now in favour of https://github.com/raspberrypi/picamera2/pull/256 which replaces it.

PR #172, opened 2022-06-15
Prefix internal functions with underscore.

IanTBlack:
This PR changes the names of functions that had a suffixed underscore to have a prefixed underscore to denote internal use. This should remove confusion when generating documentation and which function to use for those that use interactive IDEs.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -477,7 +477,7 @@ class Picamera2:
         if self.raw_index >= 0:
             self.update_stream_config(camera_config["raw"], libcamera_config.at(self.raw_index))
 
-    def configure_(self, camera_config=None) -> None:
+    def _configure(self, camera_config=None) -> None:
         """Configure the camera system with the given configuration."""
         if self.started:
             raise RuntimeError("Camera must be stopped before configuring")
@@ -542,7 +542,7 @@ class Picamera2:
 
     def configure(self, camera_config=None) -> None:
         """Configure the camera system with the given configuration."""
-        self.configure_(camera_config)
+        self._configure(camera_config)
 
     def camera_configuration(self) -> dict:
         """Return the camera configuration."""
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks for this update. Actually I would like to revisit the naming (and even the precise definition) of some of these functions. My concerns are that I regard many of these functions as being for "different audiences", but I wouldn't necessarily describe them as "internal". Of course, many will seem "internal" to some audiences.

For example, I used the convention of an underscore on the end to help show that the function is safe to call when you're operating inside the thread that is handling camera events. If you call something without an underscore, you have to be careful that you can stop it from blocking and figure out how you're going to get control back when it's finished.

To many users (those who just type commands into the main Python thread) this will see quite obscure, but to folks writing Qt applications, it will be the normal situation.

Other functions, like `dispatch_functions` seem pretty "internal", but I could easily imagine a sophisticated user wanting to perform some operation synchronously with the camera frames at which point they'd define an operation (being careful to use "underscore" functions and avoiding blocking the thread), pass it to `dispatch_functions` and then using `wait` to find out when it's done. They might have other calculations to do so you might do those before even calling `wait`.

So I'm open to ideas here. Do you know of any solutions to this that Python programmers have used? Or indeed suggest a better naming strategy to make all this clearer? I'd be very grateful for help with all this.

On a housekeeping note, development is moving to the "next" branch, as folks were getting unhappy with development happening on "main" where we would break examples for the current release. There is quite a bit of work targeted at this branch at the moment, I wish things would settle down a bit more!

IanTBlack:
[Here are some guidelines for naming conventions.](https://peps.python.org/pep-0008/#naming-conventions)

My take away is that prefixed underscores show internal use, with the added benefit of not being imported when doing import * (if a class name or separate function name). If a function or variable inside of a class has a preceding underscore, it can still be called by a user. I believe functions or variables preceded by two underscores are also not callable from inside of a class. A suffixed underscore is used by convention to avoid a conflict with a Python keyword, but it just convention.

Perhaps there could be two naming styles, one for advance uses and one for not. For example, granular functions could have `get_ or make_` in front of them and higher level functions for new users could have `capture_` which express the function use. The capture functions could be made up of the get functions. You already have something like this with `make` in the requests module..

e.g. 

`def capture_array(**inputs_here):
    buffer = get_buffer()
    np_array = make_array(buffer)
    return np_array
`

I will see about moving this to the next branch.  There are many changes I would like to suggest. I just need to make sure they don't conflict!

PR #173, opened 2022-06-16
Misc fixes

davidplowman:
Just a couple of random but very small things that I noticed while doing more substantial work elsewhere!

### DIFF
--- a/picamera2/__init__.py
+++ b/picamera2/__init__.py
@@ -1,3 +1,21 @@
 from .picamera2 import Picamera2, Preview
 from .request import CompletedRequest, MappedArray
 from .converters import YUV420_to_RGB
+
+import libcamera
+
+
+def libcamera_transforms_eq(t1, t2):
+    return t1.hflip == t2.hflip and t1.vflip == t2.vflip and t1.transpose == t2.transpose
+
+
+def libcamera_colour_spaces_eq(c1, c2):
+    return c1.primaries == c2.primaries and c1.transferFunction == c2.transferFunction and \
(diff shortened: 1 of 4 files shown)


davidplowman:
The failure recorded above in `app_test.py` is interesting. Looks rather like we're getting back a request that was made (and in fact fulfilled) before the camera was stopped, something I remember worrying about when `get_ready_requests` was made blocking.

davidplowman:
Re-running the tests makes them pass - the problem is very intermittent. Checking with the libcamera team, the problem is known about and being working on, so we may do best simply to ignore the issue until a fix turns up quite soon.

chrisruk:
Re. the intermittent issue you mention is that related to this - https://github.com/raspberrypi/picamera2/runs/7022803503?check_suite_focus=true

```
Running /home/pi/_work/picamera2/picamera2/tests/qt_gl_preview_test.py... 	ERROR
	 [69:53:41.567070685] [29845] error camera camera.cpp:528 camera in configured state trying queuerequest() requiring state running
```

Which seemed to sometimes fail for me in my github actions

davidplowman:
Ah, interesting. Yes I think it probably is. The fault described above means that a request can pop out after the camera has been stopped (which used not to happen because we would clear them out). Now we're probably trying to return it (re-queue it) as well but the camera is in its stopped (or technically, "configured") state, leading to this error.

There is a patch for this on the libcamera mailing list already. Basically the get_ready_requests is made (optionally) non-blocking again, so it would be safe to call it after stopping the camera (thereby clearing out the lurking requests again). But it sounds like we probably want to get that fix before releasing any new version.

PR #176, opened 2022-06-20
API changes for the Foundation

davidplowman:
This serious of commits implements most of the proposed changes. In particular

* The `Picamera2.xxx_configuration()` functions have become `Picamera2.create_xxx_configuration()`.
* Object-style configuration has been introduced for users who prefer this style of syntax.
* A number of functions have acquired extra optional parameters so as to reduce the necessary "boilerplate" in starting the camera.
* Controls have also acquired an optional object-style syntax.

I haven't yet looked at choosing encoders or the correct form of output object from a filename yet. There is still some auto-configuration for encoders (such as calculating a bitrate automatically) to be implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,17 @@
 
 ### Added
 
+* A very trivial Metadata class is provided that can be used to wrap the metadata dictionaries, for those that prefer this style.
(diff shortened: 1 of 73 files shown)


davidplowman:
Huh, this has failed again on that libcamera stopping problem. So we can safely ignore it here, but we do need to get their fix when they commit it.

PR #185, opened 2022-06-29
Misc fixes

davidplowman:
Just a small collection of minor fixes and improvements.

### DIFF
--- a/picamera2/__init__.py
+++ b/picamera2/__init__.py
@@ -1,3 +1,21 @@
 from .converters import YUV420_to_RGB
 from .picamera2 import Picamera2, Preview
 from .request import CompletedRequest, MappedArray
+
+import libcamera
+
+
+def libcamera_transforms_eq(t1, t2):
+    return t1.hflip == t2.hflip and t1.vflip == t2.vflip and t1.transpose == t2.transpose
+
+
+def libcamera_colour_spaces_eq(c1, c2):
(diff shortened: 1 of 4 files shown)


kbingham:
Without context...
      File "/home/pi/_work/picamera2/picamera2/tests/context_test.py", line 29, in <module>
        main()
      File "/home/pi/_work/picamera2/picamera2/tests/context_test.py", line 22, in main
        metadata = picam2.capture_file("no_context_test.jpg")
      File "/home/pi/_work/picamera2/picamera2/picamera2/picamera2.py", line 879, in capture_file
        self.capture_file_(file_output, name, format=format)
      File "/home/pi/_work/picamera2/picamera2/picamera2/picamera2.py", line 869, in capture_file_
        request.save(name, file_output, format=format)
      File "/home/pi/_work/picamera2/picamera2/picamera2/request.py", line 213, in save
        total_gain = metadata["AnalogueGain"] * metadata["DigitalGain"]
    KeyError: 'AnalogueGain'

That does indeed look like the metadata isn't set. Do you check that the Request completed successfully before this point, and that it wasn't being returned in the Cancelled state?

davidplowman:
I do check that they're completed. But, perhaps you know, I'm currently having quite a lot of trouble with requests hiding in the Python bindings layer after I stop the camera. Then they pop out when Picamera2 restarts but are "stale". There is a fix for that under discussion, and I'm speculating that this may be the same problem, though it normally manifests itself elsewhere.

PR #186, opened 2022-06-29
API changes for the Foundation

davidplowman:
This is quite a large number of commits that implements a number of additions and changes that were requested by the Foundation. We have:

* The ability optionally to use structures embedded in the Picamera2 object to configure the camera.
* The ability optionally to use objects to set control values instead of dictionaries.
* Some changes to allow the amount of boilerplate code to be reduced. For example, start() and start_recording() can optionally be given a camera configuration, and will configure the camera with default values if complete unconfigured. start() can also start a preview window.
* Video encoders can optionally be asked to calculate bitrates and quality parameters for themselves.
* There are some convenient "very high level" API functions which may suit some simple use cases for those who want to know less about what Picamera2 does ("start_and_capture_files", "start_and_record_video" etc.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,17 @@
 
 ### Added
 
+* A very trivial Metadata class is provided that can be used to wrap the metadata dictionaries, for those that prefer this style.
+* A Controls class is provided which gives an object-like view to control lists. The class is able to check that the control names are valid.
+* Configuration structures are provided so that the Picamera2 object can be configured using the "object" style. Three such instances are embedded in the Picamera2 object, namely preview_configuration, still_configuration and video_configuration.
+
 ### Changed
+
+* The start_recording() method accepts an optional configuration.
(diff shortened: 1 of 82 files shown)


davidplowman:
Yep, this failure is the classic "lurking request reappears when it shouldn't" problem.

chrisruk:
I think this all looks good to merge now! Just a few thoughts, for potential future alterations:

Should we be using the libcamera functions directly, or could we abstract them e.g. with:
```
config = picam2.create_preview_configuration(transform=libcamera.Transform(hflip=1, vflip=1))
```

I wonder if the following use of an empty dictionary might be a little confusing:
```
capture_config = picam2.create_still_configuration(raw={})
```

The easy jpeg and video functions look great!  Along with the new quality parameters for video capture.

The new configuration pattern of the following looks neat too:
```
picam2.video_configuration.size = (800, 480)
```

@davidplowman  Are you ok for me to merge now?

PR #193, opened 2022-07-02
RFC: Switch Lambda's to `partial()`

meawoppl:
I was trying to debug some things I am doing with picamera2, and this [log site](https://github.com/raspberrypi/picamera2/blob/main/picamera2/picamera2.py#L666) is substantially less useful as it just reports something like "<function <lambda> at 0x7fa8acc75a60>". This helps with that a bunch, and also is generally a bit cleaner/less code.

Heres a more extended example:
```python
In [1]:  def foo():
   ...:     pass
   ...: 

In [2]: bar = lambda: foo()

In [3]: print(bar)
<function <lambda> at 0x7fa8acc75a60>

In [4]: from functools import partial

In [5]: baz = partial(foo)

In [6]: print(baz)
functools.partial(<function foo at 0x7fa8acc75820>)
```


meawoppl:
As an aside, I think a lot of the `self.functions` -> execution -> signal -> (wait/notify/signal) -> result logic could be condensed by using a `concurrent.futures.Future` object.

davidplowman:
Hi, thanks for this. All future development is actually happening on the "next" branch now, so that we don't keep changing the examples on the "main" branch that folks refer back to. Could you target this pull request at the that branch instead? Thanks. Also if you could delete the merge commit that would make the style checker happier!

davidplowman:
I've made the suggested changes so this PR is now replaced by https://github.com/raspberrypi/picamera2/pull/235

@meawoppl Perhaps you could double-check that this is what you intended, and that you're happy for us to use your commit as per the [contribution guidelines](https://github.com/raspberrypi/picamera2#contributing). Thanks!

PR #215, opened 2022-07-08
Add convenience list of available sensor formats

will-v-pi:
Add the read_formats() function, which reads the available raw formats,
their frame rates, and other parameters from the camera, and stores them
in available_formats.

Each format is stored as a dictionary with keys: format, unpacked,
bit_depth, size, fps, crop_limits, and exposure_limits.

Signed-off-by: William Vinnicombe <william.vinnicombe@raspberrypi.com>

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -10,6 +10,7 @@ import threading
 from enum import Enum
 from typing import List
 import time
+import re
 
 import libcamera
 import numpy as np
@@ -23,6 +24,7 @@ from picamera2.utils import initialize_logger
 from .configuration import CameraConfiguration, StreamConfiguration
 from .controls import Controls
 from .request import CompletedRequest
+from .sensor_format import SensorFormat
 
 STILL = libcamera.StreamRole.StillCapture
 RAW = libcamera.StreamRole.Raw
@@ -105,6 +107,7 @@ class Picamera2:
             self.preview_configuration = self.create_preview_configuration()
             self.still_configuration = self.create_still_configuration()
             self.video_configuration = self.create_video_configuration()
+            self.read_formats()
         except Exception:
             self.log.error("Camera __init__ sequence did not complete.")
             raise RuntimeError("Camera __init__ sequence did not complete.")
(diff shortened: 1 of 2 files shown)


picamera2/picamera2.py
```diff
@@ -105,6 +107,7 @@ def __init__(self, camera_num=0, verbose_console=None, tuning=None):
             self.preview_configuration = self.create_preview_configuration()
             self.still_configuration = self.create_still_configuration()
             self.video_configuration = self.create_video_configuration()
+            self.read_formats()
```
davidplowman, line 110: I think I've changed my mind a little bit, based on what you said about all the extra log messages that come out. I think we should not call this ourselves but expose it as an extra function or property that users can access.

Maybe we could make it a property that gets filled when you first request it, and after that it's cached?

As a point of nomenclature, we tend to use "format" where we mean a combination of "Bayer order, bit depth and packing", so where we add other stuff (e.g. resolution, crop, frame rate info etc.) I'd like to call that something else, maybe "mode"? So perhaps the property becomes "sensor_modes" (that fits with sensor_format and sensor_resolution that we already have).

picamera2/picamera2.py
```diff
@@ -293,6 +296,30 @@ def open_camera(self) -> None:
         else:
             raise RuntimeError("Failed to initialize camera")
 
+    def read_formats(self) -> None:
+        """Read the available sensor formats
+        """
+        raw_config = self.camera.generate_configuration([libcamera.StreamRole.Raw])
```
davidplowman, line 302: I expect we'll change the function name in line with what I said earlier, and perhaps the comment could point out that the camera will get reconfigured by calling it!!

picamera2/picamera2.py
```diff
@@ -293,6 +296,30 @@ def open_camera(self) -> None:
...
+
+        for pix in raw_formats.pixel_formats:
+            fmt = SensorFormat(str(pix))
+            all_format = {}
+            all_format["format"] = fmt
+            all_format["unpacked"] = fmt.unpacked
+            all_format["bit_depth"] = fmt.bit_depth
+            for size in raw_formats.sizes(pix):
+                cam_format = all_format.copy()
+                cam_format["size"] = tuple([int(i) for i in str(size).split("x")])
```
davidplowman, line 314: Getting the dimensions by turning this into a string and splitting it feels slightly more awkward than I would have expected... is this a libcamera.Size object? If so it should have a width and height field.

picamera2/sensor_format.py
```diff
@@ -0,0 +1,19 @@
+import re
+
+
+class SensorFormat():
+    def __init__(self, fmt_string):
+        pixels, self.packing = fmt_string.split("_")
+        self.bit_depth = int(re.search("\d+$", pixels).group())
+        self.arrangement = re.search("[RGB]+", pixels).group()
```
davidplowman, line 8: All good here, but I find "arrangement" slightly vague... maybe "bayer_order" would be more specific?

davidplowman:
@will-v-pi Thanks for this PR. All looks great to me apart from where I've changed my mind and some minor attempts to rename things. Do you want to fix those things up and post a second version? Then we'll get that merged once I'm back. I've also enabled the auto-test-runs for you. Thanks!!

davidplowman:
This all looks great. I think the one final thing is that I'd probably prefer to squash the two commits together and to amend the commit message accordingly. I don't think the world particularly needs to know that I was dithering a bit about exactly what I wanted!

We could also add a test for this functionality. Maybe in the tests folder add something that queries the available formats and then checks it can successfully request each one using configure? But we can add that later, I'd be happy to merge this one first.

davidplowman:
Actually I've squashed them together and pushed the result directly, so closing this PR!

PR #222, opened 2022-07-18
Add full GUI app example

will-v-pi:
Add an example GUI app, with still and video capture options, along with a full set of camera controls to demonstrate the picamera2 and Qt functionality.

Requires #215 to work correctly, as it makes use of Picamera2.sensor_modes to allow selection of raw format.

Also move the app examples into a separate folder.


davidplowman:
Really liking this! I have quite a few comments, mostly nit-picks and other really annoying stuff, but also a few more general questions. In no particular order...

* I'd quite like to be able to do still captures using a "preview mode" and doing a mode switch. Possibly there could be a check box for "run the preview in the same mode as capture", and this would avoid the mode switch altogether (which is the current behaviour).
* I like the HDR feature, that's fun. I'd quite like the app to run even for people who don't install OpenCV, is there a way we could accomplish that? Have the HDR function greyed out? Perhaps in a separate tab if the current one is getting full? Or even a separate app?
* The dark background in the preview pane when the aspect ratio is wrong slightly offends my eyes, perhaps we could make the colour closer to the rest of the app?
* In the image tuning pane, I wonder about having sliders instead of numbers. It would also indicate the value you've chosen numerically too, and have a "reset" (to default) button.
* The pan/zoom is great. But I suppose making the little rectangle (that shows the crop being used) draggable might be a step too far?
* For the AGC/AEC controls it might be nice to be able to set the exposure/gain jointly. I guess that means an "Apply" button which would become enabled when you edit either of the values.
* I had some problems when enabling/disabling the AWB and changing the mode. There was some weird oscillating behaviour which would sometimes go away if I disabled/enabled it again. Might be worth investigating what's causing that, indeed if there's actually a real problem in the imaging pipeline.

will-v-pi:
Thanks, I think I've made all those changes.

> * The pan/zoom is great. But I suppose making the little rectangle (that shows the crop being used) draggable might be a step too far?

I've managed to implement this, so you can drag and scroll the rectangle to pan and zoom.

> * I had some problems when enabling/disabling the AWB and changing the mode. There was some weird oscillating behaviour which would sometimes go away if I disabled/enabled it again. Might be worth investigating what's causing that, indeed if there's actually a real problem in the imaging pipeline.

This was being caused by setting the colour gains and enabling AWB at the same time, which I've fixed, so not an issue with the pipeline.

eat-sleep-code:
As the author of [this camera software](https://github.com/eat-sleep-code/camera), I was in the process of upgrading it to take advantage of Picamera2.  

Then I stumbled upon this commit and checked out app_full.py as I figured it appeared to be a good starting spot for my application instead.

However, when I try to run it I get an error of `Picamera2' object has no attribute 'sensor_modes'.

On this particular camera, I am using [this Arducam module](https://www.uctronics.com/arducam-high-quality-camera-for-raspberry-pi-12-3mp-1-2-3-inch-imx477-hq-camera-module-with-6mm-cs-lens-for-pi-4b-3b-2b-3a-pi-zero-and-more.html).   I am not sure if this may be causing an issue, because the SD card on my camera that uses an official Raspberry Pi HQ is inside an enclosure that unfortunately would require a lot of disassembly.

Or am I possibly missing some library?

davidplowman:
@eat-sleep-code Yes, unfortunately the "official" _Picamera2_ release is missing some required features. This will be in the next release, but if you're feeling brave you could install or check out the "next" branch (warning: even more bleeding edge software!) from the repository which has the "sensor_modes" functionality.

eat-sleep-code:
When will the next release be?  That will help me decide how daring I want to be.  😝

davidplowman:
I think we're talking "a few weeks", we'll probably try and time it with an OS update, but that's just my guesstimation of the situation.

PR #228, opened 2022-07-21
Improvement to circular buffer

chrisruk:
* Make h264 repeating seq header option true by default

* Add option to circular buffer, to allow preventing frames to always be written to file


chrisruk:
Looks like the smoke tests failed because of 
```
    NameError: name 'V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER' is not defined
```
I think this might be due to the test machine not using the latest v4l2 library

davidplowman:
Closing and re-opening to force the tests to run again!

examples/capture_circular_nooutput.py
```diff
@@ -0,0 +1,21 @@
...
+dur = 5
+micro = int((1 / fps) * 1000000)
+vconfig = picam2.create_video_configuration()
+vconfig['controls']['FrameDurationLimits'] = (micro, micro)
+picam2.configure(vconfig)
+encoder = H264Encoder()
+output = CircularOutput(buffersize=fps * dur, outputtofile=False)
+output.fileoutput = "file.h264"
+picam2.start_recording(encoder, output)
+time.sleep(dur)
```
davidplowman, line 20: So this is where we don't seem to get a 5 second output, right? One thing to try might be to add
```
picam2.capture_metadata()
```
_before_ the `sleep` command (but after starting the recording). That will force it to wait for the first frame to come out of the camera before starting the timer. That will factor out any "startup time".

davidplowman:
Hmm. I seem to have failed to force the tests to run again. I'm not sure I'll ever quite understand how GitHub works!

@chrisruk One more minor question - do you want to add this example to the list of standard tests that we run? (tests/test_list.txt) Other than than it looks good to me, though it would be nice to see all the tests pass. Do you want to check the conflicts and re-push it, that might make it happen...

examples/capture_circular_nooutput.py
chrisruk: Yep, that's correct.  Will try that approach, thanks.

chrisruk:
Yep, just working on fixing the conflicts, will add to the test list too

chrisruk:
I'll update some missing docstrings for circular output.  Adding picam2.capture_metadata()
didn't seem to alter the duration of the file output though, will see if I can have a dig further into this.

davidplowman:
@chrisruk Just let me know when you'd like me to merge this. Thanks!

chrisruk:
@davidplowman I think this is ready to merge now, I found why the resulting video ended up at 3s rather than 5s, it seemed the circular buffer was getting filled, and the first frame which was a keyframe was being removed.  So the writer had the skip a number of non-keyframes, to write the first keyframe and frames after that.

By making the circular buffer slightly bigger than the time we sleep for, I now get a 5s video.

davidplowman:
Cool, always good to get to the bottom of things! Will merge now.

PR #235, opened 2022-07-26
Use functools.partial in place of lambda

davidplowman:
functools.partial makes code easier to debug and can therefore be
preferable.


davidplowman:
Closing because I've merged https://github.com/raspberrypi/picamera2/pull/250 instead.

PR #236, opened 2022-07-27
Add example and test for sensor_modes

will-v-pi:
Add an example to demonstrate the use of the sensor_modes to configure the camera, and add a test to check that it is working correctly.

### DIFF
--- a/examples/pick_mode.py
+++ b/examples/pick_mode.py
@@ -1,0 +1,37 @@
+#!/usr/bin/python3
+
+# Example of reading the available modes, and picking one with
+# the highest framerate and a raw bit depth of at least 10
+
+from picamera2 import Picamera2
+from picamera2.encoders import H264Encoder, Quality
+from picamera2.outputs import FfmpegOutput
+import time
+
+
+picam2 = Picamera2()
+
+available_modes = picam2.sensor_modes
+min_bit_depth = 10
+available_modes = list(filter(lambda x: (x["bit_depth"] >= min_bit_depth), available_modes))
+available_modes.sort(key=lambda x: x["fps"], reverse=True)
+[print(i) for i in available_modes]
+chosen_mode = available_modes[0]
+
+picam2.video_configuration = picam2.create_video_configuration(
+    raw={"size": chosen_mode["size"], "format": chosen_mode["format"].format}
+)
+picam2.configure("video")
+
+encoder = H264Encoder()
+output = FfmpegOutput('test.mp4')
+
+# Set the fps
+fps = chosen_mode["fps"]
+picam2.set_controls({"FrameRate": fps})
+
+picam2.start_recording(encoder, output, quality=Quality.LOW)
+
+time.sleep(5)
+
+picam2.stop_recording()
--- a/picamera2/sensor_format.py
+++ b/picamera2/sensor_format.py
@@ -1,4 +1,6 @@
 import re
+from libcamera import Transform
+import numpy as np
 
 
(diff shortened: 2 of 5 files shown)


examples/pick_mode.py
```diff
@@ -0,0 +1,37 @@
...
+picam2.video_configuration = picam2.create_video_configuration(
+    raw={"size": chosen_mode["size"], "format": chosen_mode["format"].format}
+)
+picam2.configure("video")
+
+encoder = H264Encoder()
+output = FfmpegOutput('test.mp4')
+
+# Set the fps
+fps = int(chosen_mode["fps"])
```
davidplowman, line 30: I was interested by the integer rounding here, but indeed it seems to be necessary. Looking at [controls.py](https://github.com/raspberrypi/picamera2/blob/main/picamera2/controls.py#L11), I use `//` to convert to convert the framerate to a duration but didn't realise that this can return a float (e.g. `1 // 1.0` returns `1.0`, not `1`). Oops.

I think it would be good to fix that up because requiring the framerate to be an integer is really not right (you couldn't ask for a frame rate of 1 frame per 2 seconds, for example)! Could I maybe ask you to add an extra first commit to this PR which resolves that problem in controls.py? Then the int cast here can be removed.

tests/mode_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+picam2 = Picamera2()
+
+for mode in picam2.sensor_modes:
+    picam2.video_configuration = picam2.create_video_configuration(
+        raw={"size": mode["size"], "format": mode["format"].format},
+    )
+    picam2.configure("video")
+    picam2.set_controls({"FrameRate": int(mode["fps"])})
+    picam2.start()
```
davidplowman, line 18: After calling `configure` I think we could retrieve the current configuration (`picam2.camera_configuration()`) and check what's in the raw stream. For each mode we could configure the camera using both the "normal" (unpacked) and packed formats.

tests/mode_test.py
```diff
@@ -0,0 +1,23 @@
...
+    # Check we got roughly the right framerate
+    print(f"Framerate: {picam2.controls.FrameRate:.2f}, requested: {mode['fps']:.2f}")
+    assert isclose(picam2.controls.FrameRate, round(mode["fps"]), abs_tol=1.0)
```
davidplowman, line 21: I'm not sure checking this really adds very much... as you'll have seen in controls.py, you're really just retrieving the same framerate that you set earlier!

Rather than recording a video, one idea might be just to run the camera for, say, 1 second and then capture the metadata. This should contain a "FrameDuration", which is actually measured from frames delivered by the camera, and should tie up with the number you requested. Of course, it requires quite a lot of stuff to be working correctly, right down to the register and timing calculations in the sensor driver!

picamera2/sensor_format.py
```diff
@@ -21,3 +21,9 @@ def unpacked(self):
 
     def __repr__(self):
         return self.format
+
+    def flip(self, v=False, h=False):
+        if h:
+            self.bayer_order = self.bayer_order[1::-1] + self.bayer_order[3:1:-1]
+        if v:
+            self.bayer_order = self.bayer_order[2:] + self.bayer_order[:2]
```
davidplowman, line 29: Just one small thing - I wonder if we might make this take a libcamera Transform object instead, as it's then fully general.  The class is documented [here](https://github.com/raspberrypi/libcamera/blob/main/src/libcamera/transform.cpp).

Issue #244, opened 2022-07-31
[BUG] Can't add overlay to DRM preview


callidus:
**Describe the bug**
If you try to set an overlay when using DRM (none-X11) preview an error is generated

**To Reproduce**
Run the `overlay_drm.py` example from the examples folder

**Expected behaviour**
Overly should work

**Console Output, Screenshots**
```
python ./overlay_drm.py 
[19:34:00.008408942] [26564]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+3700-f30ad033
[19:34:00.046462844] [26565]  INFO RPI raspberrypi.cpp:1368 Registered camera /base/soc/i2c0mux/i2c@1/imx477@1a to Unicam device /dev/media3 and ISP device /dev/media0
[19:34:00.050357758] [26564]  INFO Camera camera.cpp:1029 configuring streams: (0) 640x480-XBGR8888
[19:34:00.050808017] [26565]  INFO RPI raspberrypi.cpp:759 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected sensor format: 2028x1520-SBGGR12_1X12 - Selected unicam format: 2028x1520-pBCC
Traceback (most recent call last):
  File "/home/pi/cam_stuff/./overlay_drm.py", line 20, in <module>
    picam2.set_overlay(overlay)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/picamera2.py", line 1278, in set_overlay
    self._preview.set_overlay(overlay)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/previews/drm_preview.py", line 76, in set_overlay
    with mmap.mmap(new_fb.planes[0].fd, w * h * 4, mmap.MAP_SHARED, mmap.PROT_WRITE) as mm:
AttributeError: 'pykms.pykms.DumbFramebuffer' object has no attribute 'planes'
```

**Hardware :**
Raspberry Pi 4 with HQ camera module attached via CSI

**Additional context**
Running on the latest (at time of opening this ticket) 32bit Raspberry PI OS Lite image (no desktop).

```
pi@camera1:~/cam_stuff $ pip freeze
certifi==2020.6.20
chardet==4.0.0
colorzero==1.1
distro==1.5.0
gpiozero==1.6.2
idna==2.10
numpy==1.23.1
opencv-python==4.6.0.66
[...]

PR #246, opened 2022-07-31
Fix bug in drm_preview overlays

callidus:
was trying to access the fd via the non-existent `planes` member. Instead use the fd(plane) accessor

Fixes https://github.com/raspberrypi/picamera2/issues/244

Using a DRM preview seemed to be broken, the frame buffer object having no `planes` member. After digging about a bit in https://github.com/tomba/kmsxx/ I noticed in particular https://github.com/tomba/kmsxx/blob/2236a8ccacdfed5ff5f6873ed6618eccf570193d/kms%2B%2B/inc/kms%2B%2B/framebuffer.h#L29 this suggest that the correct way to access an fd for the desired framebuffer plane is to use `fd(plane)` rather than `planes[plane].fd`. I have confirmed this fixed the issue for me and the example `https://github.com/raspberrypi/picamera2/blob/main/examples/overlay_drm.py` now works as intended.

Signed-off-by: Tim Kelsey tkelsey@protonmail.com

### DIFF
--- a/picamera2/previews/drm_preview.py
+++ b/picamera2/previews/drm_preview.py
@@ -73,7 +73,7 @@ class DrmPreview(NullPreview):
             h, w, channels = overlay.shape
             # Should I be recycling these instead of making new ones all the time?
             new_fb = pykms.DumbFramebuffer(self.card, w, h, "AB24")
-            with mmap.mmap(new_fb.planes[0].fd, w * h * 4, mmap.MAP_SHARED, mmap.PROT_WRITE) as mm:
+            with mmap.mmap(new_fb.fd(0), w * h * 4, mmap.MAP_SHARED, mmap.PROT_WRITE) as mm:
                 mm.write(np.ascontiguousarray(overlay).data)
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks for submitting this. Indeed there seems to have been a pykms change that I didn't notice.

The first change (line 76) seems good but I'm getting an error with the change at line 128. As far as I can tell, the original code is still correct there. Could you maybe double-check that?

Could you also maybe just check that your signed-off-by is in the commit message as well as the PR description, I wasn't quite sure whether it was.

Thanks!

davidplowman:
I've committed a version of this that just makes the first change but not the second (https://github.com/raspberrypi/picamera2/pull/249). I believe that's enough to resolve the issue, but if not, please re-open it or send another PR.

Thanks for your help in letting us know about the problem!

callidus:
Humm, my GH notifications were not setup properly so I missed this. Sorry for not responding, glad the fix was useful :+1:

Issue #313, opened 2022-09-12
[BUG] UDP messages can be too long, aren't broken up


CorvetteCole:
**Describe the bug**
When utilizing UDP streaming as described in the docs, a resolution, bitrate, etc higher than shown easily cause an OSError: [Errno 90] Message too long. This appears to be because the UDP packets are larger than allowed by the OS, and need to be split up.

**To Reproduce**
Use a resolution and bitrate large enough to create big packets. Example script (must change target ip):
```
#!/usr/bin/python3

import socket
import time

from picamera2 import Picamera2
from picamera2.encoders import H264Encoder
from picamera2.outputs import FileOutput

picam2 = Picamera2()
video_config = picam2.create_video_configuration({"size": (1640, 1232)})
picam2.configure(video_config)
encoder = H264Encoder(1000000, repeat=True)

with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
    sock.connect(("10.11.9.112", 1234))
    stream = sock.makefile("wb")
    picam2.start_recording(encoder, FileOutput(stream))
    try:
        time.sleep(10000)
    finally:
        picam2.stop_recording()
```

**Expected behaviour**
No crashes because of UDP packets that are too big.

**Console Output, Screenshots**
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/encoders/v4l2_encoder.py", line 190, in thread_poll
    self.outputframe(b, keyframe, (buf.timestamp.secs * 1000000) + buf.timestamp.usecs)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/encoders/encoder.py", line 183, in outputframe
    out.outputframe(frame, keyframe, timestamp)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/outputs/fileoutput.py", line 77, in outputframe
    self._write(frame, timestamp)
  File "/home/pi/.local/lib/python3.9/site-packages/picamera2/outputs/fileoutput.py", line 95, in _write
    self._fileoutput.write(frame)
  File "/usr/lib/python3.9/socket.py", line 722, in write
    return self._sock.send(b)
OSError: [Errno 90] Message too long
```

**Hardware :**
[...]

PR #314, opened 2022-09-13
add socketoutput, capable of splitting packets for datagram socket

CorvetteCole:
This has not been tested yet. Basically, we initialize this very similarly to the FileOutput, but there is one important difference. The socket must be open before this output is started. This is intended to solve https://github.com/raspberrypi/picamera2/issues/313

### DIFF
--- a/picamera2/outputs/socketoutput.py
+++ b/picamera2/outputs/socketoutput.py
@@ -11,7 +11,7 @@ class SocketOutput(Output):
     """Socket handling functionality for encoders"""
 
     def __init__(self, socket=None, pts=None):
-        """Initialise UDP output
+        """Initialise socket output
 
         :param socket: Socket to write frames to, defaults to None
         :type socket: socket.socket, optional
@@ -39,7 +39,7 @@ class SocketOutput(Output):
             self._socket = None
         elif isinstance(socket, socket.socket):
             if self._socket is not None:
-                self._socket.close()
+                self.close()
             self._datagram = socket.type == socket.SOCK_DGRAM
             self._socket = socket
 
@@ -65,7 +65,7 @@ class SocketOutput(Output):
             raise RuntimeError("Must pass callback function or None")
 
     def outputframe(self, frame, keyframe=True, timestamp=None):
-        """Write frames to a UDP stream
+        """Write frames to a socket stream
 
         :param frame: Frame
         :type frame: bytes
--- a/picamera2/outputs/socketoutput.py
+++ b/picamera2/outputs/socketoutput.py
@@ -1,0 +1,111 @@
+"""Write frames to a socket"""
+
+import types
+
+from .output import Output
+
+MAX_DATAGRAM_SIZE = 65507
+
+
+class SocketOutput(Output):
+    """Socket handling functionality for encoders"""
+
+    def __init__(self, socket=None, pts=None):
+        """Initialise UDP output
+
+        :param socket: Socket to write frames to, defaults to None
+        :type socket: socket.socket, optional
+        :param pts: File to write timestamps to, defaults to None
+        :type pts: str or BufferedWriter, optional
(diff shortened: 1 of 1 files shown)


davidplowman:
Thanks for posting this.

@chrisruk Is this the kind of thing you think we're looking for?

My initial observation was that the new class might be duplicating a lot of the FileOutput code? Could we derive it from FileOutput instead?

Obviously there would be some further work in tidying the commits, adding some tests and so on, but that is secondary at this point. @CorvetteCole You might want to check our [contribution guidelines](https://github.com/raspberrypi/picamera2#contributing) if you haven't done so, I don't think there's anything surprising or scary in them, but always best to be on the same page!

Thanks!

chrisruk:
Yeah I think it makes sense to extend or modify the FileOutput class.

I did just think we could have something like a split parameter in FileOutput constructor, which could default to 0, for no splitting of data. Then if you set it to 65507, it would split a frame into chunks of 65507 or less.

chrisruk:
Yeah that would be nice, I had a little look into if we could do that, but it seems .makefile creates something like a bufferedreader, so you can't detect that it's got a UDP socket behind it as far as I can tell.

davidplowman:
I think you can do
```
stream = sock.makefile("wb")
print(stream.raw._sock)
```
Bit ugly, though!!

chrisruk:
Cool, didn't know you could do that.  Which would mean we could potentially modify FileOutput to something like this -

```
    def _write(self, frame, timestamp=None):
        try:
            if isinstance(self._fileoutput.raw, socket.SocketIO) and \
                self._fileoutput.raw._sock.type == socket.SocketKind.SOCK_DGRAM:
                # split data here, calling self._fileoutput.write on each chunk
            else:
                self._fileoutput.write(frame)
```

davidplowman:
Indeed, I'd be slightly tempted to do the detection only once in the constructor (and set a `split` variable/property), but otherwise yes!

CorvetteCole:
I agree, this class is very much a duplicate of FileOutput. A split parameter in FileOutput sounds pretty nice for preventing duplicate code.

I personally am in favor of automatically detecting datagram streams, although we would have to be careful in case a passed in file object is not a socket type and doesn't have the _sock attribute. I guess we can wrap that in a try except, thoughts?

The alternative may be to implement the SocketOutput as a subclass of FileOutput and then override some methods, if it would be desirable to separate those. 

Ultimately I ask these questions because I am a little unclear on the architectural goals around these outputs. The purpose of this SocketOutput was _mostly_ just intended to spark some discussion.

Side note: I have read the contribution guidelines... thanks! I've been working super closely integrating this project, so I'd be remiss not to contribute back :)

chrisruk:
> I personally am in favor of automatically detecting datagram streams, although we would have to be careful in case a passed in file object is not a socket type and doesn't have the _sock attribute.

Good point, I believe the following line should then mean _sock must be present-
```
isinstance(self._fileoutput.raw, socket.SocketIO)
```
Python does also have hasattr() to check an attribute is present.

Also that's a good point @davidplowman re. doing the checking in the constructor.

So I think it should be possible to make FileOutput do the splitting of data for us, if a datagram socket is passed even though
it's abstracted somewhat via .makefile(), which means the existing UDP example would work for higher bitrates/resolutions.

CorvetteCole:
I'll see if I can give this a go on my test setup Friday. Will leave this PR open until I or one of you file a new PR for the revised FileOutput. Alternatively, I might just overwrite this PR's contents entirely.

chrisruk:
Thanks everyone for the comments, I've just created a revised FileOutput here - https://github.com/raspberrypi/picamera2/pull/317

davidplowman:
Thanks for all the discussion on this one - I'll close it now in favour of Chris's new PR!

PR #317, opened 2022-09-15
Splitting of frame into multiple UDP datagrams

chrisruk:

### DIFF
--- a/picamera2/outputs/fileoutput.py
+++ b/picamera2/outputs/fileoutput.py
@@ -1,5 +1,7 @@
 """Writes frames to a file"""
 
+import io
+import socket
 import types
 
 from .output import Output
@@ -31,14 +33,20 @@ class FileOutput(Output):
     @fileoutput.setter
     def fileoutput(self, file):
         """Change file to output frames to"""
+        self._split = False
         self._firstframe = True
         if file is None:
             self._fileoutput = None
         else:
             if isinstance(file, str):
                 self._fileoutput = open(file, "wb")
-            else:
+            elif isinstance(file, io.BufferedWriter):
                 self._fileoutput = file
+            else:
+                raise RuntimeError("Must pass io.BufferedWriter")
+            if hasattr(self._fileoutput, "raw") and isinstance(self._fileoutput.raw, socket.SocketIO) and \
+                    self._fileoutput.raw._sock.type == socket.SocketKind.SOCK_DGRAM:
+                self._split = True
 
     @property
     def connectiondead(self):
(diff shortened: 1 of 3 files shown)


chrisruk:
I've tested that the splitting algorithm functions locally, by just generating a random binary file and splitting it and concatenating the splits and running a checksum on it.  My network is wifi based, which isn't working well with UDP at the moment though.

picamera2/outputs/fileoutput.py
```diff
@@ -31,14 +33,20 @@ def fileoutput(self):
     @fileoutput.setter
     def fileoutput(self, file):
```
davidplowman, line 34: Maybe there's a small enhancement we could make here.

If we accept a `split` parameter (defaulting to `None`) then the caller could explicitly supply an integer which would be the size we split packets into. Below, where we detect the UDP socket we could change the `None` into 65507.

The benefit would be that users can choose a size that is better for their own network, for example they might know that there's an MTU of 1500 that gets enforced somewhere, and so this might behave better for them.

davidplowman:
One other thing might be to add a second commit that simply adds a test to the tests/test_list.txt script. It would ensure we don't ever accidentally break this and get bitten again by it!

The examples/capture_stream_udp.py script will very nearly do, we would just need to increase the bitrate (by quite a lot), and possibly the resolution (not sure) to guarantee that it will tickle the big-packet problem.

picamera2/outputs/fileoutput.py
chrisruk: Yep, that's a good point, I'll add a split=None, to the FileOutput constructor.

I assume we only need to make use of the split value for UDP packets, as there wouldn't 
be any advantage for TCP etc. sockets?

davidplowman: Yes, I think that's right, TCP will sort itself out.

CorvetteCole:
This looks great, I'll test this on our problematic setup tomorrow.

chrisruk:
@CorvetteCole Cool, thanks!

@davidplowman I just added a test case, which when run for the current 'next' branch without these changes produces

```
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/pi/Repositories/picamera2/picamera2/encoders/v4l2_encoder.py", line 190, in thread_poll
    self.outputframe(b, keyframe, (buf.timestamp.secs * 1000000) + buf.timestamp.usecs)
  File "/home/pi/Repositories/picamera2/picamera2/encoders/encoder.py", line 183, in outputframe
    out.outputframe(frame, keyframe, timestamp)
  File "/home/pi/Repositories/picamera2/picamera2/outputs/fileoutput.py", line 77, in outputframe
    self._write(frame, timestamp)
  File "/home/pi/Repositories/picamera2/picamera2/outputs/fileoutput.py", line 95, in _write
    self._fileoutput.write(frame)
  File "/usr/lib/python3.9/socket.py", line 722, in write
    return self._sock.send(b)
OSError: [Errno 90] Message too long
```

davidplowman:
@chrisruk It's all looking good to me! Do you want to merge it?

chrisruk:
:) Yes please

CorvetteCole:
This works on our problematic setup, and fixes the issue we were seeing

davidplowman:
Great, thanks for the confirmation!

Issue #324, opened 2022-09-18
[BUG] capture_image bombs out when specifying "lores" instead of "main" as argument


tvoverbeek:
**Describe the bug**
Capturing a PIL image only works for the "main" stream.
Taking the example in section 6.1.2 of the manual but replacing "main" with "lores"
causes an error and when used in a program  hangs it because the image request is never completed

**To Reproduce**
Use the following code:
```
from picamera2 import Picamera2
import time
picam2 = Picamera2()
picam2.start()
time.sleep(1)
image = picam2.capture_image("lores")
```
Here is a console log:
```
tvo@picam2-32:~ $ python bug.py
[4:07:33.004749969] [2906]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+3866-0c55e522
[4:07:33.031900844] [2907]  INFO RPI raspberrypi.cpp:1374 Registered camera /base/soc/i2c0mux/i2c@1/ov5647@36 to Unicam device /dev/media1 and ISP device /dev/media0
[4:07:33.035236964] [2906]  INFO Camera camera.cpp:1035 configuring streams: (0) 640x480-XBGR8888
[4:07:33.035597290] [2907]  INFO RPI raspberrypi.cpp:761 Sensor: /base/soc/i2c0mux/i2c@1/ov5647@36 - Selected sensor format: 640x480-SGRBG10_1X10 - Selected unicam format: 640x480-pgAA
Traceback (most recent call last):
  File "/home/tvo/bug.py", line 6, in <module>
    image = picam2.capture_image("lores")
  File "/usr/lib/python3/dist-packages/picamera2/picamera2.py", line 1259, in capture_image
    return self._execute_or_dispatch(partial(self.capture_image_, name), wait, signal_function)
  File "/usr/lib/python3/dist-packages/picamera2/picamera2.py", line 1038, in _execute_or_dispatch
    done, result = function()
  File "/usr/lib/python3/dist-packages/picamera2/picamera2.py", line 1243, in capture_image_
    result = request.make_image(name)
  File "/usr/lib/python3/dist-packages/picamera2/request.py", line 150, in make_image
    return self.picam2.helpers.make_image(self.make_buffer(name), self.picam2.camera_config[name], width, height)
  File "/usr/lib/python3/dist-packages/picamera2/request.py", line 134, in make_buffer
    with _MappedBuffer(self, name) as b:
  File "/usr/lib/python3/dist-packages/picamera2/request.py", line 17, in __init__
[...]

PR #327, opened 2022-09-20
Improvements to error handling

davidplowman:
A couple of small but helpful commits:

The first tidies up better after an error, so that the system doesn't get stuck in a "you haven't waited for the last thing to finish" state.

The second adds more intelligible messages to a couple of reasonably common failure cases.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -992,8 +992,12 @@ class Picamera2:
 
     def wait(self):
         """Wait for the event loop to finish an operation and signal us."""
-        result = self._future.result()
-        self._future = None
+        try:
+            result = self._future.result()
+        except Exception as e:
+            raise
+        finally:
+            self._future = None
         return result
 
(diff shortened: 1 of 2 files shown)


davidplowman:
@chrisuk I think this is a good change and does seem to resolve the problems in https://github.com/raspberrypi/picamera2/issues/324. Do you want to take a look and merge if you're happy? No great urgency though, we're only talking about the next update release, which is probably a small number of weeks away. Thanks!

picamera2/picamera2.py
```diff
@@ -992,8 +992,12 @@ def process_requests(self) -> None:
 
     def wait(self):
         """Wait for the event loop to finish an operation and signal us."""
-        result = self._future.result()
-        self._future = None
+        try:
+            result = self._future.result()
+        except Exception as e:
+            raise e
```
chrisruk, line 998: Small question, does 'raise e' behave the same as 'raise' which is used lower down? (I assume it probably does?)

PR #329, opened 2022-09-21
Initial changes for raw video recording

chrisruk:

### DIFF
--- a/examples/capture_video_raw.py
+++ b/examples/capture_video_raw.py
@@ -1,0 +1,52 @@
+#!/usr/bin/python3
+import time
+
+import numpy as np
+from picamera2 import Picamera2
+from picamera2.encoders import Encoder
+from pidng.core import RAW2DNG, DNGTags, Tag
+from pidng.defs import *
+from PIL import Image
+
+size = (2592, 1944)
+picam2 = Picamera2()
+video_config = picam2.create_video_configuration(raw={"format": 'SGBRG10', 'size': size})
+picam2.configure(video_config)
+picam2.encode_stream_name = "raw"
+encoder = Encoder()
+
+picam2.start_recording(encoder, 'test.raw', pts='timestamp.txt')
+time.sleep(5)
+picam2.stop_recording()
+
+buf = open("test.raw", "rb").read(size[0] * size[1] * 2)
+arr = np.frombuffer(buf, dtype=np.uint16).reshape((size[1], size[0]))
+
+# Scale 10 bit / channel to 8 bit per channel
+im = Image.fromarray((arr * ((2**8 - 1) / (2**10 - 1))).astype(np.uint8))
+im.save("test-8bit.tif")
+
+# Note - this will look very dark, because it's 10 bit colour depth of image, in
+# a 16 bit / channel tiff
+im2 = Image.fromarray(arr)
+im2.save("test-16bit.tif")
+
+# Create DNG file from frame, based on https://github.com/schoolpost/PiDNG/blob/master/examples/raw2dng.py
+# Tested loading of DNG in darktable
+r = RAW2DNG()
+t = DNGTags()
+bpp = 10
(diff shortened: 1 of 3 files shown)


chrisruk:
It might be necessary to move the is_raw functions to their own python file, as I can't use Picamera2.is_raw without getting a circular import alas.  Maybe there's another solution here though.

I need to validate the data output to the .raw file too.

davidplowman:
Looks good to me, but I see the awkwardness about setting the format. There are probably several options, but that doesn't help knowing which one to choose!

1. We could largely ignore the format. As you say, I think only derived classes look at it.
2. The V4L2 format values could possibly go down into the V4L2 encoder? There are V4L2 enums for all the raw formats, but I don't see any benefit in actually deciphering them all. Though there's still the question of validating the format in the Encoder class (unless we ignore it!).
3. We also have a SensorFormat class in sensor_format.py - does that help? It doesn't exhaustively catch everything that's "bad", but we could always beef it up a bit if we needed.

I'm slightly liking option 1 at the moment... which is what you currently have!

chrisruk:
Yeah I think option 1 probably makes sense.

I just altered the example to extract a frame from the raw data.  Initially I thought I might need to use stride rather than width, but width seemed to be what worked for me here?

I made it generate two tiffs, one scaled from 10 bit --> 8 bit, and one left as 16 bit, with the 10 bit data inside.

chrisruk:
@davidplowman I think this is ready for review now.  It outputs the first frame pre-demosaicing as tiffs and a DNG in the example.

davidplowman:
Hi Chris, yes LGTM. The only thing maybe might be to rearrange the commits a bit so that we have one commit that makes the encoder changes, and a second that adds the example (probably squash those example changes into one)? Apart from that I'm good to merge. Thanks!

chrisruk:
Thanks David good point, I've just modified the commits.   So the example file should only appear in the 2nd commit now.  Hopefully they're in order now.

Out of interest, is this how you would have removed the example file from the first commit (which is the approach I just took) - 
```
git rebase -i 14c3fac0fef718e156eb34c09c0acd5983880211^  # commit you want to edit, change pick to edit in interactive editor
git rm capture_video_raw.py 
git commit --amend
git add capture_video_raw.py 
git rebase --continue
```

tf4ong:
@chrisruk  when will this be pushed to the main code? I couldn't run it with the pip installed version. It is giving me RuntimeError: Invalid format

chrisruk:
I'm not sure when it will be in 'main' afraid, but it should be in the 'next' branch soon.  Yeah you'll get an Invalid Format issue if you run with the PyPi/deb installed version, as I made changes to part of the Encoder class, which aren't in those versions.

davidplowman:
@chrisruk Hi Chris, commits look good now (the way you did it looks better than what I would probably have done!!), but if I might be a real pain for a moment, could you maybe improve the commit messages? The 2nd commit that adds the example is particularly cryptic... (obviously I realise it's just the message from the second of the two squashed commits). But then I'll merge it, promise!!

@tf4ong We'll get this pushed onto "next" very shortly. You can install that branch from Github using pip (using some special syntax), though I would recommend uninstalling it again once we update the apt package. Installing via apt is generally preferable because if you install via pip there's a risk that libcamera and Picamera2 can get out of sync. An alternative to installing from pip is just to clone that branch from the repo, and then put /home/pi/picamera2 (or wherever you cloned it) into your PYTHONPATH environment variable.

chrisruk:
@davidplowman Good point re. commit messages, think they're a bit more descriptive now.  Took me a while to figure out I needed to add --no-verify with git commit --amend (otherwise the hook didn't seem to let me amend the commit message).

tf4ong:
@chrisruk @davidplowman Many thanks guys will give it a try. Really awesome that you guys made such an amazing package. Its leaps ahead for scientific imaging compared to previous module

Issue #307, opened 2022-09-09
[BUG] Library throwing DEBUG log msgs, even when set for INFO Level.


ajonnet:
Debug level logs are being thrown by the library, irrespective of what value is supplied for "verbose_console" param for constructor Picamera2(). It happens when log handlers are defined prior to calling of constructor.

Due to above behaviour, Logs are filled up with debug statements being generated by the library. 

**To Reproduce**
import logging
from picamera2 import Picamera2

logger = logging.getLogger("")
formatStr = '[%(asctime)s] [%(threadName)s] [%(levelname)s]: %(message)s'
console = logging.StreamHandler()
console.setFormatter(logging.Formatter(formatStr))
logger.addHandler(console)
logger.setLevel(logging.INFO)
picam=Picamera2(verbose_console=1)

**Console Output**
[0:33:57.988721016] [91]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+3866-0c55e522
[0:33:58.091540556] [95]  WARN RPI raspberrypi.cpp:1258 Mismatch between Unicam and CamHelper for embedded data usage!
[0:33:58.093436522] [95]  INFO RPI raspberrypi.cpp:1374 Registered camera /base/soc/i2c0mux/i2c@1/imx219@10 to Unicam device /dev/media3 and ISP device /dev/media0
2022-09-09T06:32:59.722Z | INFO     | Initialization successful.
[2022-09-09 12:02:59,722] [MainThread] [INFO]: Initialization successful.
2022-09-09T06:32:59.723Z | INFO     | Camera now open.
[2022-09-09 12:02:59,723] [MainThread] [INFO]: Camera now open.
[2022-09-09 12:02:59,723] [MainThread] [DEBUG]: <libcamera._libcamera.CameraManager object at 0x7f90beda30>

**Expected behaviour**
As verbose_console is set to 1, below debug statement should not have generated
[2022-09-09 11:34:32,216] [MainThread] [DEBUG]: <libcamera._libcamera.CameraManager object at 0x7fae3a4bb0>

**Hardware :**
Pi 3B
OS: BullsEye
Package: python3-picamera2
Version: 0.3.3-1

To avoid the issue, below patch works for now (to be called after initialization of Picamera2)
logger = logging.getLogger('picamera2')
logger.setLevel(logging.INFO)

davidplowman:
Hi, thanks for raising this. Unfortunately I don't know much about the Python logging module, but here's what I found out.

[...]

PR #332, opened 2022-09-27
Revised logging

davidplowman:
These commits address https://github.com/raspberrypi/picamera2/issues/307.

The first commit revises the way the logging system is used according to the proposal I made at the bottom of that discussion.

One snag is that if you don't configure any logging at all, the default behaviour is for warning and error messages to be output, but without the WARNING or ERROR tags at the start of the line. So for the scripts in our tests folder, I've added a 2nd commit which configures the logging (by calling `Picamera2.set_logging`) so that these tags appear, and the automated testing would then flag them up.

I'd still like the chance to get more feedback if there is any, so no merging yet please!


tvoverbeek:
There is problem with the static set_logging function. I am in the hospital right now. Will comment when I am back home. Probably at the end of this week.

chrisruk:
I was just wondering if you do - picam2.set_logging(logging.ERROR), should that also not alter what logging we receive from libcamera, or not?  Or are we treating them as separate logs?

tvoverbeek:
As far as I know libcamera logging is independent from picamra2, so picamera2 should set the corresponding environment variables for libcamera.
What these are remain TBD.

chrisruk:
Yeah, I just found you can do the following in __init__
```
os.environ["LIBCAMERA_LOG_LEVELS"]="*:ERROR"
```
But in set_logging it would be too late to affect some of the messages printed from libcamera.  Would it possibly be an idea to be able to set logging level in __init__ and then we could also change libcameras env variable

davidplowman:
Personally I'm against doing anything with the libcamera logging at the moment. It's a different system, and doesn't look the same. Ideally it would be nice to get Python bindings to control it. I think it's fine for us to revisit at some point, for now I'd like to sort out the Picamera2 end of things!

Anyway, I'm happy to wait with all this until next week, thanks to everyone for their help in the meantime!

tvoverbeek:
The problem is in lines 69-70:
```
        while _logger.hasHandlers():
            _logger.removeHandler(_logger.handlers[0])
```
In case the user has already created a logger but before the picamera2 _logger has created a handler,
`_logger.hasHandlers()` will return True since the parent logger has a handler.
But in case no _logger handlers have been defined yet`_logger,handlers[0]` does not exist => crash.
Found this code which would avoid this:
```
        for handler in _logger.handlers:
            if isinstance(handler, logging.StreamHandler):
                _logger.removeHandler(handler)
``

**Comment by davidplowman** [comment]
Thanks for sending that. The logging module sure has more pitfalls than I ever imagined!!

Just one thought, however. Is it safe to remove something from a list that is being iterated over? Might this be better:
```
        while _logger.handlers:
            _logger.removeHandler(_logger.handlers[0])
```

**Comment by tvoverbeek** [comment]
Should be equivalent, but removing an item from the list while iterating is OK.

**Comment by davidplowman** [comment]
Hi everyone, I've made the necessary changes and de-WIP-ed this PR. If anyone has any final suggestions...    Thanks!

**Comment by tvoverbeek** [comment]
@davidplowman No further suggestions from me, except the logging should be documented (did not check if something is already there)

**Comment by chrisruk** [comment]
Looks good to me too.  Very minor thing in one place you use logging.WARN rather than Picamera2.WARN

**Comment by davidplowman** [comment]
> Looks good to me too. Very minor thing in one place you use logging.WARN rather than Picamera2.WARN

Yes, I don't know if it's worth the both of changing it. The only reason `Picamera2.WARN` (for example) exists is so that users don't have to `import logging` in order to call the `set_logging` method with one of the levels. I don't know if there's a better way to achieve that?

**Comment by chrisruk** [comment]
Gotcha, makes sense re. saving importing logging.  Want me to merge this now?

**Comment by davidplowman** [comment]
I guess so...!

**Comment by chrisruk** [comment]
I just found you can apparently do .clear() to clear a list of things -

```
_logger.handlers.clear()
```

(https://stackoverflow.com/questions/7484454/removing-handlers-from-pythons-logging-loggers)

davidplowman:
Are we confident that the `removeHandler` call doesn't do anything else that the simple call to `clear` might be missing?

davidplowman:
Ah, I've found the source code [here](https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L1713). So I guess not.  Seems like it's a good change then!

chrisruk:
Yeah was just on that myself, seems to just acquire a lock and remove

chrisruk:
Thanks for the change.  Github action doesn't seem to be running at the moment strangely

davidplowman:
Closing in order to re-open and see if the automated tests run again!

PR #346, opened 2022-10-08
Async Call Cleanup

meawoppl:
There are a number of functions designed to be call by the camera event loop which are dispatched asynchronously. They all return things of the form `(True, return_value)` so I pulled out the boolean, as that always evaluates true, this also highlighted some bound locals that didn't need to be passed into closured functions explicitly, and results in generally much more readable internals.

Related work:
* added some typing information (Moved to #358)
* Added position indication to lint errors. (Moved to #359)

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -8,7 +8,7 @@ import selectors
 import tempfile
 import threading
 from enum import Enum
-from typing import List
+from typing import List, Tuple, Any, Dict, Optional
 import time
 from concurrent.futures import Future
 from functools import partial
@@ -151,7 +151,7 @@ class Picamera2:
         self._encoder = None
         self.pre_callback = None
         self.post_callback = None
-        self.completed_requests = []
+        self.completed_requests: List[CompletedRequest] = []
         self.lock = threading.Lock()  # protects the functions and completed_requests fields
         self.have_event_loop = False
         self.camera_properties_ = {}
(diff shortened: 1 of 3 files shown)


picamera2/picamera2.py
```diff
@@ -1035,10 +1034,9 @@ def _execute_or_dispatch(self, function, wait, signal_function):
                 raise RuntimeError("Failure to wait for previous operation to finish!")
             self._dispatch_functions([function], signal_function)
             if self.completed_requests:
-                done, result = function()
-                if done:
-                    self.functions = []
-                    self._future.set_result(result)
+                result = function()
+                self.functions = []
+                self._future.set_result(result)
```
meawoppl, line 1043: Note for @davidplowman. If you want the behaviors here to match the other execution point, it might make sense to wrap the exceptional case into  a future here too? Not totally clear to me.

meawoppl:
Gentle poke for thoughts @davidplowman

davidplowman:
Hi, thanks for submitting this. Would it be possible to split the various addressed here into separate PRs in accordance with our [contribution guidelines](https://github.com/raspberrypi/picamera2#contributing)? That way we could get some of the straightforward issues merged more quickly and leave us free to discuss other things without them being mixed up with other changes. Thanks!

meawoppl:
I can do that. I usually just add typing as I need to get my IDE type-checks to come up happy. GIven the diff size of <100 LOC I figured this wouldn't matter terribly. I will reopen it as ~3PRs later today.

meawoppl:
@davidplowman when the two above go in, I will rebase this one.

meawoppl:
Moved to #375

PR #349, opened 2022-10-10
Initial changes for supporting other cameras

chrisruk:


chrisruk:
@davidplowman Would you mind having a go at running this pull request on a compute module sometime, with 2 cameras (no rush!).

On my pi with a webcam + picamera, I can now get 2 preview windows running simultaneously :)

```
#!/usr/bin/python3

import time

import numpy as np

picam2a = Picamera2(0)
picam2a.configure(picam2a.create_preview_configuration())
picam2a.start_preview(Preview.QT)

picam2b = Picamera2(1)
picam2b.configure(picam2b.create_preview_configuration())
picam2b.start_preview(Preview.QT)

picam2a.start()
picam2b.start()
time.sleep(5)

overlay = np.zeros((300, 400, 4), dtype=np.uint8)
overlay[:150, 200:] = (255, 0, 0, 64)
overlay[150:, :200] = (0, 255, 0, 64)
overlay[150:, 200:] = (0, 0, 255, 64)

picam2a.set_overlay(overlay)
picam2b.set_overlay(overlay)
time.sleep(5)
```

I'm using Qt rather than QtGL here, as I made a simple change to convert MJPEG->RGB.

davidplowman:
@chrisruk Hi Chris, just tried this on my CM4 and it looks really promising!

I found all the previews worked fine together, including the QTGL one (not quite clear to me how access to the GPU gets arbitrated there, but perhaps that's a question for later on!). The only funny thing, which I see you're looking at already, is that closing one preview seemed to close them all (and leave things a bit messed up).

It's probably a bit early to review, but I did have a snoop through some of the changes. I guess the os.pipe was an easy way to signal the individual cameras in the same way and continue to use the QSocketNotifier. In some ways that seems over-complicated, but OTOH the architecture we have (where Qt is optional) really requires a Qt-free way to send signals, so maybe it needs to be like this? Don't know if there are any alternatives there.

I also thought it might tidy things a bit if the CameraManager class perhaps had register/unregister functions to tell it that it needs to watch for a particular camera (or not any more).

Thanks also for looking at the Qt threading issues. When I'm not working with Qt I tend to forget what a pain it can be! I haven't really understood all that yet, but I wonder if there's a way we could make a single thread that does all the Qt stuff and actually persists forever. We've had a long-standing problem where if you close and re-create a Qt preview, it complains about being in the wrong thread (though it works fine), and maybe there could be a route to fixing that. Not sure. Arguably we could do something like that as a separate commit even before adding multi-camera support.

Finally, I thought the MJPEG camera changes probably want to go into a separate PR - I feel we could possibly review and merge all that first? How does that work, by the way, do you get reasonable framerates with all that JPEG decode going on?

Anyway, I'm encouraged how well things work already, perhaps let's talk next week!

chrisruk:
Cool, thanks for checking it out on the CM4 :)

Yeah, that's a good point about whether we need to use QSocketNotifier, will try and have a think about that.  I think libcamera uses an eventfd rather than std. file handle, like I'm currently doing, I don't think there's builtin python support for eventfds but can look for a library for that.

> I also thought it might tidy things a bit if the CameraManager class perhaps had register/unregister functions to tell it that it needs to watch for a particular camera (or not any more)."

Good point, I do wonder, but need to confirm, if the CameraManager class for a particular process would only be seeing data for the camera IDs that are passed as cookies to create_request.

> I wonder if there's a way we could make a single thread that does all the Qt stuff and actually persists forever. We've had a long-standing problem where if you close and re-create a Qt preview, it complains about being in the wrong thread (though it works fine)

That's a good point, I've also just bumped into that message when altering the .stop() function.  I'm wondering if something is possibly persisting outside of the thread itself somehow, will see if I can dig into that.

> Finally, I thought the MJPEG camera changes probably want to go into a separate PR - I feel we could possibly review and merge all that first? How does that work, by the way, do you get reasonable framerates with all that JPEG decode going on?

Yeah thats seems a good idea re. using a separate pull request for that.  I've not analysed the fps, but it did seem to look visually ok to me.  I had to make it so that certain libcamera controls that weren't available for my webcam can be bypassed.  The webcam itself seems to just pass MJPEG data directly, so I just made a function to convert the MJPEG to RGB for Qt.

> Anyway, I'm encouraged how well things work already, perhaps let's talk next week!

Sounds good

chrisruk:
Just moved to https://github.com/raspberrypi/picamera2/pull/376, to rebase again from next, with multiple new changes

PR #356, opened 2022-10-17
Fix framerate calculation in preparation for libcamera update

davidplowman:
The next version of libcamera has a "None" default for the framerate so the code previously here will fall over. We may as well fix this ahead of time as the new code is in fact slightly preferable anyway.


chrisruk:
Am I right in thinking cam_mode["fps"], would be None by default, so you're just overwriting it?

chrisruk:
Cool, thanks

PR #357, opened 2022-10-17
Fix DRM overlays when using fkms

davidplowman:
This is part of the fix for getting display overlays to work when using DRM with fkms (rather than kms). A Linux kernel update is also required, but this change will be needed too so that the code will run correctly.


chrisruk:
Just wondering, there's not a specific exception that can be caught, rather than RuntimeError?

davidplowman:
Indeed, good question. I've trailed it down and the error actually comes from [here](https://github.com/tomba/kmsxx/blob/master/kms%2B%2B/src/atomicreq.cpp#L68). So I think `RuntimeError` is as good as it gets. There's probably a way to find out exactly what properties the display driver supports, but catching the error works for me!!

chrisruk:
Cheers, thanks for the code link. Will merge now

PR #358, opened 2022-10-17
Typing Updates

meawoppl:
This is only type annotations and the addition of a couple of docstrings.

This is split 1/3 of #83 .


meawoppl:
Blocks https://github.com/raspberrypi/picamera2/pull/346

davidplowman:
Looks good, but could you target this at the next branch as well, please? Thanks!

PR #359, opened 2022-10-17
Small Upgrade to `checkstyle.py`

meawoppl:
The style checker has some unusual requirements on spacing around type annotations, so I added this to help future developers understand where it wanted spaces.

Part 2/3 from #83 

Signed off by Matthew Goodman (matt@exclosure.io)

### DIFF
--- a/tools/checkstyle.py
+++ b/tools/checkstyle.py
@@ -346,10 +346,11 @@ class StyleChecker(metaclass=ClassRegistry):
 
 
 class StyleIssue(object):
-    def __init__(self, line_number, line, msg):
+    def __init__(self, line_number, line, msg, offset=None):
         self.line_number = line_number
         self.line = line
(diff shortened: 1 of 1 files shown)


meawoppl:
Blocks https://github.com/raspberrypi/picamera2/pull/346

davidplowman:
Hi, thanks for this one. I take it that it doesn't change what the style checker complains about, it only causes more information to be printed out when it does complain (in some circumstances, at least). Is that right?

Just one thing, could you maybe add another line or change the commit message so that it gives some indication what it's about, rather than just "small upgrade"? With that change I think I'm good to merge this. Thanks!

davidplowman:
Thanks for this patch! I've updated the commit message and merged it directly onto "next" branch.

meawoppl:
Thank you!

PR #360, opened 2022-10-18
Typing Only

meawoppl:


davidplowman:
Thanks for these patches! I've updated the commit message slightly and merged it directly.

PR #375, opened 2022-10-25
Async Cleanup

meawoppl:
From #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,28 +40,8 @@ sudo apt upgrade
 If you have installed _Picamera2_ previously using `pip`, then you should also uninstall this (`pip3 uninstall picamera2`).
 
 Thereafter, you can install _Picamera2_ _with_ all the GUI (_Qt_ and _OpenGL_) dependencies using
-
 ```
-sudo apt install -y python3-pyqt5 python3-opengl
-```
-(No changes are required to _Picamera2_ itself.)
-
-### Installation using `apt`
-
-`apt` is the recommended way of installing and updating _Picamera2_.
-
-If _Picamera2_ is already installed, you can update it with `sudo apt install -y python3-picamera2`, or as part of a full system update (for example, `sudo apt upgrade`).
-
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, so at the moment I feel that I still want to keep the ability for functions executed in that event loop to say "I need to run again". I'm planning (and actually always have been) to use it once libcamera gives us some better tools for synchronising when control changes happen. I think I gave an example of this somewhere, but the plan is that you'd be able to do things like
```
id = picam2.set_controls(control_list)
request_where_it_has_happened = picam2.capture_request(sync_id=id)
```
and the second call here would wait for libcamera to tell us that those control changes have now happened. There will also be the ability to pass in a list of control-lists and it will work through them all as fast as it can giving you one frame with each set applied. This is all still dependent on what happens within libcamera itself - we've made some proposals to the team but of course it's their decision how they want to take things forward!

meawoppl:
This makes sense. What are your feelings on making that an exception that is explicitly handled and requeued? It does seem to an exceptional case compared to a majority (all) of the 20 or so calls here.

davidplowman:
That sounds like it might be an idea. What exactly did you have in mind? Maybe it would be worth making a PR with just that one change so that we could take a look and consider merging it. If you could make sure it's done on the current top of the "next" branch (I tried merging one of these PRs previously and got into a bit of a mess doing manual conflict resolutions). Thanks!

meawoppl:
Can do. Just to clarify, you want PR's up to date base on the current state of the "next" branch generally? Is the goal here to keep the `main` branch consistent with the most recent release?

davidplowman:
Yes, "main" is supposed to correspond exactly to the current release, and whenever we make a new release we copy "next" on top of "main".

meawoppl:
That is a little unusual.... In my experience people typically use `main` as the current dev target then do release tags, which y'all are doing also? *confused*

This works, but a bunch of the structure of github will assume that `main` is the current target (will open PR's or help you merge against that ect. 

/shrug

Moved this all to #387

PR #376, opened 2022-10-25
Multiple camera support

chrisruk:


chrisruk:
To use it, you'd need to make sure you checkout this pull req/branch and set pythonpath to point to it, before running the example

davidplowman:
I've just started trying this out! One thing I noticed was that it no longer frees the camera resources like it used to - hence why I added https://github.com/raspberrypi/picamera2/pull/377. I guess there's something in the new CameraManager that is hanging on to stuff?

I also seem to have one other test failure, but let me have a snoop around first to see if I can see what's going on or make a simpler test case. Thanks!

chrisruk:
Thanks for the test case, will merge that to next and I'll rebase this pull request and try it out.

davidplowman:
I think I have some understanding of the other failure case I have. It's `tests/mode_test.py` which fails intermittently for me (using imx477). It fails on the first run often enough, though it can sometimes take many runs to fail too. The problem I see is that libcamera underneath us complaining about not being able to queue a request back to the camera, and shortly after that something else asserts and we abort.

The basic cause, I believe, is that we're sometimes hanging on to a request which we weren't previously. If we don't drop all our handles then libcamera won't throw them away early enough and we can end up trying to re-queue them later when they're no longer valid. The change in question is that we can now hold on to requests in the `_requests` field which we didn't do previously.

A "simple" fix, and which I think largely confirms this, seems to be to add `self._requests = []` to the bottom of the `stop_` method, though this needs some more thought. The previous arrangement would dispatch the `stop_` call into the `process_requests` method to run, where reading from the socket, fetching any requests and dropping everything is all synchronous and safe.

But we now have another thread spinning round, and while the "simple" fix works, I think the theoretical possibility still exists that an extra request could get dumped on the `_requests` queue even after `stop_` has returned. So possibly a needs a little more thought about what the best approach might be.

chrisruk:
Ah, that's very interesting, I don't think I've had that particular test failing for me from what I can recall, so will try and run it repeatedly till it fails.  Your description and potential fix are both very helpful!  I'll have a think about that too.

I've just been using your close test script, which was very useful!.  Just been trying a patch for that.

picamera2/picamera2.py
```diff
@@ -44,6 +46,66 @@ class Preview(Enum):
     QTGL = 3
 
 
+class CameraManager:
+    def __init__(self):
+        self.cameras = {}
+        self.setup()
+        self._lock = threading.Lock()
```
davidplowman, line 53: Looks just a bit strange to create the lock after `setup()` starts the thread that uses it. Or should we avoid calling `setup()` here because it will presumably happen when the first camera gets `add`ed?

picamera2/picamera2.py
```diff
@@ -44,6 +46,66 @@ class Preview(Enum):
...
+            events = sel.select(0.2)
+            for key, _ in events:
+                callback = key.data
+                callback()
+
+        sel.unregister(self.cms.event_fd)
+        self.cms = None
+
+    def handle_request(self, flushid=None):
+        """Handle requests
```
davidplowman, line 92: Just wondering whether the `flushid` is worth the trouble, given that we have to clear the `picam2._requests` in the `stop_()` call anyway?

picamera2/picamera2.py
```diff
@@ -148,8 +211,15 @@ def __init__(self, camera_num=0, verbose_console=None, tuning=None):
                 os.environ["LIBCAMERA_RPI_TUNING_FILE"] = tuning_file.name
         else:
             os.environ.pop("LIBCAMERA_RPI_TUNING_FILE", None)  # Use default tuning
-        self.camera_manager = libcamera.CameraManager.singleton()
+        if not Picamera2._cm:
+            Picamera2._cm = CameraManager()
+        self.notifyme_r, notifymew = os.pipe()
```
davidplowman, line 216: I wonder if we should be closing these file descriptors in our close method? I know a process can have lots of open file descriptors so it probably doesn't matter too much, but my sense of tidiness is feeling slightly uncomfortable...

picamera2/picamera2.py
```diff
@@ -148,8 +211,15 @@ def __init__(self, camera_num=0, verbose_console=None, tuning=None):
...
+        self.notifyme_rr = os.fdopen(self.notifyme_r, 'rb')
+        self.notifyme_w = os.fdopen(notifymew, 'wb')
```
davidplowman, line 218: I felt slightly that `self.notifyme_w` should be the file descriptor from `os.pipe()`. That would name things a bit more consistently with `self.notifyme_r` which is also the file descriptor. I didn't quite get the name `notifyme_rr` - why `rr` and is there something more obvious?

Also I'm not sure it's worth doing an `fdopen` on the write file descriptor, because `handle_request()` can just use `os.write(self.cameras[c].notifyme_w, b"\x00")` which would do the "flush" automatically. If that all makes sense....!

picamera2/picamera2.py
```diff
@@ -455,9 +532,9 @@ def stop_preview(self) -> None:
         """
         if self._preview:
             try:
+                self.have_event_loop = False
```
davidplowman, line 535: Was there a reason for moving this line up? Doesn't matter, I was just curious!

picamera2/picamera2.py
```diff
@@ -964,16 +1036,13 @@ def stop_(self, request=None) -> None:
...
             # up when the camera is started the next time.
-            sel = selectors.DefaultSelector()
-            sel.register(self.camera_manager.event_fd, selectors.EVENT_READ)
-            events = sel.select(0)
-            if events:
-                self.camera_manager.get_ready_requests()
-
+            self._cm.handle_request(self.camera_idx)
             self.started = False
+            with self._requestslock:
```
davidplowman, line 1044: I guess that taking the lock is probably optional here, given that we've ensured the camera can't complete any more requests and we've read out any that were still pending. Still, it's harmless!

picamera2/picamera2.py
```diff
@@ -1061,7 +1126,7 @@ def process_requests(self) -> None:
             # immediately back to libcamera. You could consider customising this number.
             # When there's only one buffer in total, don't hang on to anything as it would stall
             # the pipeline completely.
-            max_len = 0 if self.camera_config['buffer_count'] == 1 else 1
+            max_len = 0 if (not self.camera_config or self.camera_config['buffer_count'] == 1) else 1
```
davidplowman, line 1129: Slightly curious about this one. Is it suggesting we've closed the camera but are still somehow in the event loop? Which perhaps suggests we failed to stop it, though maybe that could be related to the issues we had in the `stop_()` method. Maybe at some point I might undo this and see what breaks...!

picamera2/picamera2.py
chrisruk: Good point, will avoid calling setup there

picamera2/previews/q_gl_picamera2.py
```diff
@@ -357,18 +358,20 @@ def repaint(self, completed_request, update_viewport=False):
 
     @pyqtSlot()
     def handle_requests(self):
+        self.picamera2.notifyme_rr.read(1)
```
davidplowman, line 361: I wonder if we should read everything? Just in case we ever get multiple frames without "catching up" in between.

picamera2/previews/q_picamera2.py
```diff
@@ -144,10 +144,10 @@ def resizeEvent(self, event):
 
     @pyqtSlot()
     def handle_requests(self):
+        self.picamera2.notifyme_rr.read(1)
```
davidplowman, line 147: Here too, maybe read everything?

picamera2/previews/q_gl_picamera2.py
chrisruk: I agree, definitely makes sense

picamera2/picamera2.py
davidplowman: Yes. I think that looks good!

picamera2/previews/q_gl_picamera2.py
chrisruk: Just having issues with using .read() without 1, 

In the docs it says -
"size is an optional numeric argument. When size is omitted or negative, the entire contents of the file will be read and returned"

So I assume it might try and read till EOF, but with a pipe I don't think it'd be getting that?

davidplowman: I think with a pipe you just get everything that is there - it won't wait for the pipe to close or anything like that!

picamera2/picamera2.py
chrisruk: I might have been testing something after setting it, but can't recall afraid.

picamera2/picamera2.py
chrisruk: Yeah, I'm not sure, I added it just to save creating new CompletedRequests for that camera ID

picamera2/picamera2.py
chrisruk: Good point, I'll try removing this after the latest changes.

picamera2/picamera2.py
chrisruk: Also doing the following too
```
            self.notifymeread.close()
            os.close(self.notifyme_w)
```

picamera2/previews/q_gl_picamera2.py
chrisruk: I just created a little test case, it never seems to print read interestingly -

```
import threading
import os
import time

notifyme_r, notifyme_w = os.pipe()
notifymeread = os.fdopen(notifyme_r, 'rb')

def writer():
    while True:
        os.write(notifyme_w, b"\x00")
        time.sleep(1)

def reader():
    while True:
        notifymeread.read()
        print("read")

thread1 = threading.Thread(target=writer, daemon=True)
thread1.start()

thread2 = threading.Thread(target=reader, daemon=True)
thread2.start()

time.sleep(10)
```

picamera2/picamera2.py
chrisruk: I've now reverted this change, after testing locally.

picamera2/picamera2.py
chrisruk: It seems trickier than I thought to call self.notifymeread.close() without issues.

Currently getting -
QSocketNotifier: Invalid socket 3 and type 'Read', disabling...

Which I'm just digging into.

davidplowman:
Other than the points listed above (I see some of them are still fighting back!) I've been trying out the latest version and it seems to work well for me. I also have a patch for the DRM preview which I'll post once we've merged this PR.

picamera2/previews/q_gl_picamera2.py
chrisruk: Managed to solve this using a non blocking pipe

picamera2/picamera2.py
chrisruk: Managed to solve this as far as I understand by adding - parent, preview, tup = (None, None, None) to the MonitorThread

chrisruk:
Cool, sounds good!

Just pushed a new revision, where I now am able to close the pipe.

And do read(), rather than read(1), by using a non-blocking pipe

chrisruk:
Just rebased again off next, so that it will run through the app_full test

davidplowman:
Hi Chris, I've just done a little bit of commit-tidying, pulling out a couple of bits that seemed like they could be removed from the big main commit, and filled in a longer commit message ([here](https://github.com/davidplowman/picamera2/commits/multiple-cameras)). Just want to give that one last try on my CM4 and then I'll merge it, code should be exactly the same.

davidplowman:
I've merged this now from that branch with the slightly tidied up commits. So I'll close this one now!

PR #377, opened 2022-10-26
Test that Picamera2 objects free resources properly when closed

davidplowman:
After calling close() on the Picamera2 object we start the camera again in another process - which should work!


chrisruk:
Just wondering, would it also be worth testing running again in the same process too?

davidplowman:
Good thought, I'll add some variations on closing/re-opening in the same process.

PR #378, opened 2022-10-26
Remove workaround required because get_ready_requests() was blocking

davidplowman:
For quite a while the camera manager get_ready_requests() function was blocking which meant that, when stopping the camera, we had to check if there was anything there to read out and flush before calling it. The new non-blocking arrangement is much more convenient.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -964,14 +964,9 @@ class Picamera2:
             self.stop_count += 1
             self.camera.stop()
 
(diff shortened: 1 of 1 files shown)


davidplowman:
Now that this code looks a bit less strange, it's also clearer that it's not really the right thing to do in the multi-camera world!

davidplowman:
Looking at the latest version of the multiple camera PR, I think that takes care of this change too so we can probably just close this one!

PR #386, opened 2022-10-31
App full fixes

davidplowman:
A couple of small commits:

1. Fix some bit-rot in apps/app_full.py
2. Add it to the list of automated tests so that it can't break again like that.

### DIFF
--- a/apps/app_full.py
+++ b/apps/app_full.py
@@ -670,10 +670,10 @@ class AECTab(QWidget):
     def reset(self):
         self.aec_check.setChecked(True)
         self.awb_check.setChecked(True)
-        self.exposure_time.setValue(picam2.camera_controls["ExposureTime"][2])
-        self.analogue_gain.setValue(picam2.camera_controls["AnalogueGain"][2])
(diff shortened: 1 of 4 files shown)


davidplowman:
The style checker can be ignored here - it's complaining about a file that is actually a symbolic link!

chrisruk:
Is it not complaining about whitespace/newline in the symlink file, or am I misunderstanding?

davidplowman:
It's a bit hard to read, isn't it? So I think the file `apps/app_full.py` is fine, that got updated in the first of the two commits with no complaints.

For the 2nd commit I typed `ln -s ../apps/app_full.py` so as to give me a local version of the file but which would stay synced with the one in apps, and which I could import easily to make a standalone test. So I think it's complaining about that `../apps/app_full.py` text that I wrote and which is now in symbolic link file?

chrisruk:
Oh I see heh, so with -

```
#1: : E226 missing whitespace around arithmetic operator
+../apps/app_full.py
   ^
```

It's complaining about the slash in the pathname of the symlink.  I should it was complaining about something in the file itself.

chrisruk:
Will merge now

PR #387, opened 2022-10-31
Request Queue Rework

meawoppl:
New version of #83 and #83 as per requests of @davidplowman.

This PR makes the request flow of picamera2 much simpler by:
- Making the function queue 1:1 with a futures queue.
- Removing the "completed_requests" management
- Treating  `wait` and`signal_function` the same everywhere.

This is done by using  a `deque` primitive and using builtin behaviors of `Future` objects.  Calls into the async underbelly now queue and unspool cleanly, and multiple callers into the module are supported.

### DIFF
--- a/picamera2/exceptions.py
+++ b/picamera2/exceptions.py
@@ -1,0 +1,3 @@
+class RequeueException(Exception):
+    def __init__(self):
+        super("This method needs to be run again")
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1023,7 +1023,8 @@ class Picamera2:
 
         if func:
             req = requests.pop(0)
-            self.log.debug(f"Execute function: {func} on {req}")
+            # TODO reset to debug
+            self.log.info(f"Execute function: {func} on {req}")
             try:
                 result = func(req)
                 fut.set_result(result)
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1,16 +1,16 @@
 #!/usr/bin/python3
 """picamera2 main class"""
 
-
 import json
 import os
 import selectors
 import tempfile
 import threading
 from enum import Enum
-from typing import List, Tuple, Any, Dict, Optional
+from typing import Callable, List, Tuple, Any, Dict, Optional, Deque
 import time
 from concurrent.futures import Future
+from collections import deque
 from functools import partial
 import logging
 import sys
@@ -25,6 +25,7 @@ from picamera2.previews import DrmPreview, NullPreview, QtGlPreview, QtPreview
 
 from .configuration import CameraConfiguration, StreamConfiguration
 from .controls import Controls
+from .exceptions import RequeueException
 from .request import CompletedRequest, Helpers
 from .sensor_format import SensorFormat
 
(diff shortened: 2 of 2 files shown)


davidplowman:
Hi, thanks for this PR. There are quite a few things going on here, and I was wondering if each change could be put into its own commit so that it's a bit easier to tell what has changed and why. Maybe some of them would be better in their own PRs? I'd really like to merge some of your work, but I'd really also like to take it one change at a time. For example:

* The changes to handling `wait` and `signal_function` - seems like a good thing so I'd like to be able to merge this without having to understand and merge everything else.
* Removing the `completed_requests` queue - this was fulfilling a particular function so I'm not entirely clear why this has gone. Maybe we need to think about this one some more?
* Creating a `futures` queue - I recall we talked about my use-case where I want to be able to wait for particular controls to take effect, so I'm keen to improve the code whilst still fulfilling these requirements. I haven't understood if these changes address that or not. Is there a way forward, do you think?

Thanks!

meawoppl:
Its kind of tricky to pull these things apart. I keep trying to do smaller bits, but getting a bit hungup on the entanglement of part in the `process_requests` loop which has at least 7 distinct responsibilities.

Its hard to fix the `signal_function` and `wait` semantics as they hinge on how the futures mature/lock unlock, the "residual" completed requests make it really hard to work against the other futures surface while still supporting the pluripotent call structure.

As an aside, having to rebase/squash/sign for any change (when squash-merge is a GH supported merge option), interact with a custom linter, custom testing stack, and atypical branching structure make the work per-PR very high for me.

davidplowman:
Hi again, I got the impression that  the wait/signal_function changes were maybe something that could be done in isolation, separate from everything else? It seemed like a good change, and something that could be targeted first?

As things stand I'm a bit reluctant to get rid of the completed_requests list as it performs a specific function (it reduces the time application loops have to sit waiting for the next frame after they have processed the previous one), Hence why I'm always keen to take things in pieces, easier ones first!!

meawoppl:
This makes sense, but also the completed_requests internal storage leadds to a situation where a call to `capture_frame` can return an image from before the call was issued. I think this is a pretty unexpected behavior....

davidplowman:
> This makes sense, but also the `completed_requests` internal storage leads to a situation where a call to `capture_***` can return an image from before the call was issued.

Yes that's true, maybe both mechanisms should be possible?

> Also, there is no way to ensure the capture of serial frames presently...
> 
> I think this is a pretty unexpected behavior for a camera library....

meawoppl:
The thing that is really at the lynch-pin of this complexity is [this call](https://github.com/raspberrypi/picamera2/blob/main/picamera2/picamera2.py#L1082). 
There is a very complex combinatoric possibility space that functions entertains:
- `wait` = True/False/None (3 bits) 
- `signal_function` is None or a callback (1 bit)
- `completed_requests` is empty/full (thread linked!) (1 bit)

Depending on those inputs:
* 16/32 of branches will actually run the passed in function twice, once in the thread loop, and once in the calling thread. (I think this is a bug)
* 24/32 will return the result of the function, the rest return `None`
* 8/32 will raise exceptions from the function call if it raises any

Also, I should note that it calls to `wait()` aren't safe as another call to it, or an error in the thread loop could have removed the object.

meawoppl:
@davidplowman I wanted to make sure you saw my note about how `_exectute_or_dispatch()` will often run functions twice.

davidplowman:
Thanks for the reminder, I must have missed that previously!

Could you maybe just explain in a little more detail? Looking back at that piece of code, it looks like the function will remain on the list and get called again if `done` came back as `False` (I wasn't sure if that was the case you were referring to), but I believe is the intention there. But if I've overlooked something, then please do let me know. Thank you very much for your help!

PR #388, opened 2022-10-31
Cleanup in `NullPreview`

meawoppl:
It is pretty easy for the current sync issues in picamera2 to bubble during the null preview and generally get lost forever.

This PR adds logging for the exceptional cases and corrects the event sync.

As a side note, the way that `picam2` is passed around in here is more than a little odd....

### DIFF
--- a/picamera2/previews/null_preview.py
+++ b/picamera2/previews/null_preview.py
@@ -1,6 +1,10 @@
 """Null preview"""
 
+from logging import getLogger
 import threading
+
+
+_log = getLogger(__name__)
 
 
 class NullPreview:
@@ -16,13 +20,11 @@ class NullPreview:
 
         sel = selectors.DefaultSelector()
         sel.register(picam2.camera_manager.event_fd, selectors.EVENT_READ, self.handle_request)
-        self.event.set()
+        self._started.set()
 
-        while self.running:
-            events = sel.select(0.2)
-            for key, _ in events:
-                callback = key.data
-                callback(picam2)
+        while not self._abort.is_set():
+            for _ in sel.select(0.2):
+                self.handle_request(picam2)
 
     def __init__(self, x=None, y=None, width=None, height=None, transform=None):
         """Initialise null preview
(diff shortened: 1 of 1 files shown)


picamera2/previews/null_preview.py
```diff
@@ -16,13 +20,11 @@ def thread_func(self, picam2):
 
         sel = selectors.DefaultSelector()
         sel.register(picam2.camera_manager.event_fd, selectors.EVENT_READ, self.handle_request)
-        self.event.set()
+        self._started.set()
```
meawoppl, line 23: Previously this was only ever `.set()`, so if this preview were restarted, the `.start()` method would roll forward as the event was set from the previous call to `.start()`

davidplowman:
Hi, thanks for this PR. In principle I'd be happy to merge most of this, but with a few small changes.

1. Could you fix the style checker complaints, please? I'm generally happy to ignore it when we don't like what it says, but it would be helpful just to mention that in the comments so that I know!
2. As you've probably gathered, I really don't like mixing fixes for different things together in a single commit. For example, I see that the `event` (which you've called `started_`) should be cleared when the preview starts. That's a straight and simple bug - could it go into a separate commit?
3. I wasn't sure why the `running` flag needed to be changed to an `abort_` event. Generally I try to avoid changes that aren't necessary because it makes for more stuff to look at and review, but if we should have this change then it would be nice to have it in its own commit too.

Thanks!

meawoppl:
Happy to fix the lint.

I came here trying to understand why the start/stop cycles take so long. The lack of clearing the `_started` event (the underscore is the way that python people [mark methods not intended for public use](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name#:~:text=Python%20doesn't%20have%20real,not%20part%20of%20the%20API.)) was certainly a bug, and can lead to subtle bugs when starting/stopping the camera. 

I renamed the `running` to `_abort` for a couple reasons:
- I was already changing the variable to a `threading.Event` [here is a related SO post](https://stackoverflow.com/a/43879462).
- To be consistent with the naming/typing of other thread communication primitive
- To avoid being mistaken for an external signal as to whether or not the preview is running (which it certainly isn't)

davidplowman:
HI again, so I propose to close this PR in favour of https://github.com/raspberrypi/picamera2/pull/392 which is hopefully equivalent - could you take a look and confirm that you're happy that I've marked them as being signed off by you? Thanks!

PR #389, opened 2022-11-01
Support for `RequestCopy`

meawoppl:
Rework the `CompletedRequest` class to add a simple `.copy()` class method that returns a object with the same call structure that isn't backed by a `libcamera.request` object. This allows users to keep the useful `save_*` and `as_*` methods without holding a reference to a request object and keep the stream moving.

This required the following:
- Movement of `is_***()` methods to a top level module named `formats.py` to resolve circular import issues.
- Making a common ABC that both `CompletedRequest` and `RequestCopy` derive from

### DIFF
--- a/picamera2/encoders/encoder.py
+++ b/picamera2/encoders/encoder.py
@@ -122,9 +122,6 @@ class Encoder:
             if type(self) is Encoder:
                 """Currently allow any format that is
                 passed, when Encoder used directly.
-                We can't use Picamera2.is_raw(value) etc.
-                to validate formats here, as we get a circular import
-                when trying to import Picamera2
                 """
                 self._format = value
             else:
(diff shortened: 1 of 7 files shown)


picamera2/encoders/encoder.py
```diff
@@ -122,9 +122,6 @@ def format(self, value):
             if type(self) is Encoder:
                 """Currently allow any format that is
                 passed, when Encoder used directly.
-                We can't use Picamera2.is_raw(value) etc.
-                to validate formats here, as we get a circular import
-                when trying to import Picamera2
```
meawoppl, line 127: No longer true ,but out of scope for now.

davidplowman:
Hi, thanks for this PR. I must confess I don't know what these changes do and why they are helpful. Could you perhaps explain? Thanks.

meawoppl:
As an aside, the way `picam2` is passed into basically every object in this repository is a pretty strong anti-pattern in the OO world, this is referred to as a ["god object"](https://en.wikipedia.org/wiki/God_object). 

Its hard to refactor right now, in a bunch of places because of this. One example of why this is fraught is that presently a `CompletedRequest` will reference the mutable object `picam2.camera_config` in several places to learn what the length/width/format is, but it could have changed since the class was instantiated, leading to unpredictable behavior/crashes.

PR #391, opened 2022-11-01
Memory improvements in app_full.py

davidplowman:
We reduce the size of the lores stream that we allocate for stills captures. Otherwise it defaults to the full resolution which is way bigger than we need, and causes some significant problems:

* It allocates more memory than necessary, which is always at risk of failure.

* The lores images can be too large for the GPU to display on Pi 3 or earlier devices - this would cause an immediate crash.


chrisruk:
Cheers, gotcha

PR #394, opened 2022-11-02
encoders import tidy.

meawoppl:
Star imports are kinda sketchy. In this case, the `ctypes` module was getting imported through the v4l2 module...

This can lead to circular imports and other confusion. I can see what that module does not use an `__all__` block it looks like it was codegened :/

### DIFF
--- a/picamera2/encoders/h264_encoder.py
+++ b/picamera2/encoders/h264_encoder.py
@@ -1,10 +1,11 @@
 """H264 encoder functionality"""
 
 from math import sqrt
-from picamera2.encoders import Quality
-from picamera2.encoders.v4l2_encoder import V4L2Encoder
-from v4l2 import *
 
+from v4l2 import (V4L2_CID_MPEG_VIDEO_H264_I_PERIOD,
+                  V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER, V4L2_PIX_FMT_H264)
+
+from picamera2.encoders import Quality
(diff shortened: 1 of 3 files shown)


davidplowman:
Thanks for this PR. In principle l think we should merge this but could you edit the commit message to say what you did, with your sign-off below it? Then someone can simply hit the big merge button. Thanks!

@chrisruk Are you OK with this one?

chrisruk:
Yeah, this looks a good idea, just wondering, do you mind running isort on the changed files (which orders the imports and will also remove duplicate imports, which I just noticed we have some of).

```
pip install isort
```

meawoppl:
Frankly I don't understand this groups approach to lint. I suggested in some previous PR's how to standardize that this all (#94), but it instead seems to be interesting to y'all to consider what the best answer on each PR individually in addition to having a custom lint tool that diverges from industry standards...

meawoppl:
Files sorted as per request.

meawoppl:
@chrisruk was there anything else here?

chrisruk:
Thank you for fixing the PR.

We'd like to merge it but would need a commit message that 
corresponds to our guidelines, please see - https://github.com/raspberrypi/picamera2#Contributing

meawoppl:
@davidplowman @chrisruk when you have a chance.

davidplowman:
Thanks for the update. The commit message is currently saying "Lower indentation" which wouldn't really be correct, could you update it to say something like "Tidy encoder imports" or something like that? Then we're good to merge. Thanks!

PR #395, opened 2022-11-02
Dead File

meawoppl:
Remove a file with a single invalid import statement...

### DIFF
--- a/picamera2/utils/__init__.py
+++ b/picamera2/utils/__init__.py
@@ -1,1 +1,0 @@
-from .picamera2_logger import initialize_logger


davidplowman:
Thanks for this PR. I'd like to merge it but it would be very helpful if you could add a commit message which says what you did (all I'm seeing is "Signed off by ...") and then add your sign-off below that. With that I could just push the "merge" button and the process would be quicker. I have tried to explain this in the[ contribution guidelines](https://github.com/raspberrypi/picamera2#contributing), but please ask if anything there is not clear. Thanks!

meawoppl:
@davidplowman updated as per your request.

davidplowman:
Thank you!

PR #398, opened 2022-11-02
Reflow Init Logic

meawoppl:
Some small work to make the init logic more clear and concise.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -399,31 +399,31 @@ class Picamera2:
         :return: True if success
         :rtype: bool
         """
-        if self.camera_manager.cameras:
-            self.camera = self._grab_camera(self.camera_idx)
-        else:
+        if not self.camera_manager.cameras:
             self.log.error("Camera(s) not found (Do not forget to disable legacy camera with raspi-config).")
             raise RuntimeError("Camera(s) not found (Do not forget to disable legacy camera with raspi-config).")
 
-        if self.camera is not None:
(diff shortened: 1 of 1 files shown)


chrisruk:
Just noticed the smoke tests seem to be failing here, you can run the tests locally from the repo doing the following with X running -

```
python3.9 tools/run_tests.py -p $(pwd)
```

And there's also a git hook mentioned in the readme, that will run the style checker which is rather handy

meawoppl:
In all honesty, I have been using the runner attached to this repo as a sort of CI system. I am setting up a small rack of pi4b's later next week with various attached cameras, but until then, opening PR's is the easiest way to run tests.

meawoppl:
@chrisruk also, the linter has somewhat inconsistent behaviour compared to my local run. It seems to want slightly unusual spacing around type hints specifically compared to the usual associated PEP8 style.

meawoppl:
@chrisruk I fixed the lint/style issues as well as the logic that was preventing startup. Its probably out of scope for this change, but the (non-zero) return code for`camera.acquire()` tells whether there camera isn't connected or is held by another process, which seems very useful from a user perspective.

meawoppl:
@chrisruk Conflict resolved, updated to the current `next` commit changed to match request.

meawoppl:
@chrisruk @davidplowman Need anything else here?

davidplowman:
Thanks for the updates. I've merged this directly now with just a small update to the commit message.

PR #400, opened 2022-11-02
Multicam drm preview

davidplowman:
The first commit enables two DRM preview windows to be made when multiple cameras are available, one for each camera.

The second commit adds a simple test for this.


chrisruk:
Hi David, will merge this soon as managed to try this on a pi4 with a webcam :), by converting MJPEG->RGB.

Would a pull request for MJPEG DRM support be ok?  (would need to tidy up what I tried)

PR #402, opened 2022-11-03
MJPEG support for DRM

chrisruk:
Note that this is slower because we convert MJPEG
to RGBA.

### DIFF
--- a/tests/test_list_drm.txt
+++ b/tests/test_list_drm.txt
@@ -1,0 +1,2 @@
+examples/overlay_drm.py
+examples/preview_drm.py
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import subprocess
 import sys
+import time
 
 # Execute this script with no arguments to run all the tests listed in the
 # tests/test-list.txt file. It prints out whether each test passes, fails,
@@ -65,8 +66,16 @@ def print_subprocess_output(exc: subprocess.CalledProcessError):
         print(indent(exc.stderr.decode('utf-8')))
 
 
-def run_tests(tests):
+def run_tests(tests, xserver=True):
     """Run all the given tests. Return the number that fail."""
+    if not xserver:
+        vt = 1
+    else:
+        vt = 7
+    if os.system(f"sudo chvt {vt}") != 0:
+        print("FAILED to switch VT")
+        return len(tests)
+    time.sleep(3)
     num_failed = 0
     for test in tests:
         clean_directory()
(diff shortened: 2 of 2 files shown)


chrisruk:
I was wondering if there is a way to switch from X to the console programmatically, so that DRM could be tested too?

At the moment I just use ctrl-alt-f1 / f7 to switch between the two.

chrisruk:
I was wondering if there is a way to switch from X to the console programmatically, so that DRM could be tested too?

At the moment I just use ctrl-alt-f1 / f7 to switch between the two.

davidplowman:
Apparently (just found this out) you can use `sudo chvt 1`, and then `sudo chvt 7` to return. So it would indeed by good to run the DRM tests automatically using this...

picamera2/previews/drm_preview.py
```diff
@@ -49,13 +48,17 @@ class DrmPreview(NullPreview):
         "XBGR8888": pykms.PixelFormat.XBGR8888,
         "YUV420": pykms.PixelFormat.YUV420,
         "YVU420": pykms.PixelFormat.YVU420,
+        "MJPEG": pykms.PixelFormat.XBGR8888,
```
davidplowman, line 51: I think we can simplify things a little if we map MJPEG onto `pykms.PixelFormat.BGR888`. We must also make a change below....

picamera2/request.py
```diff
@@ -203,7 +203,11 @@ def make_array(self, buffer, config):
             # cv2.cvtColor(image, cv2.COLOR_YUV2BGR_YUYV) will convert directly to RGB.
             image = array.reshape(h, stride // 2, 2)
         elif fmt == "MJPEG":
-            image = np.array(Image.open(io.BytesIO(array)))
+            image = Image.open(io.BytesIO(array))
+            if drm:
+                image = image.convert('RGBA').tobytes()
+            else:
+                image = np.array(img)
```
davidplowman, line 210: With those other tweaks I think we can revert this change, and then also remove the extra `drm` parameter, leaving things a bit simpler.

picamera2/previews/drm_preview.py
chrisruk: Thanks, that seems a good idea

davidplowman:
I see those DRM tests are not happy... the test machine is headless, maybe that's what makes the difference?

chrisruk:
Yeah it could well be because it's headless, as I do check the return code from the call.

Just added a sleep statement in case that makes a difference.

chrisruk:
Actually just pulled the HDMI cable from my display, and also get 'RuntimeError: Failed to reserve DRM plane'

davidplowman:
I think you can run a "fake" display which might do what we want. Try putting `video=HDMI-A-1:1920x1080@60D` on the end of your /boot/cmdline.txt file and reboot. Does that work at all?

chrisruk:
Yes, that seemed to work.

The only difference I noticed was some printed messages after running examples/overlay_drm.py that said
```
commit failed
commit failed
```

davidplowman:
Yes, I think so. Do all the DRM examples produce messages like that? And is it just a few, or is it an "every frame" kind of thing?

davidplowman:
And you'll need a micro-to-regular HDMI adaptor for that dummy plug!! :)

chrisruk:
Ah, I worked out why I got that message, it prints it when headless, if you're on the Xserver VT.

So if you do 'sudo chvt 1' and then run the overlay_drm.py demo, it works fine :)

Don't really need the hdmi dongles now heh.

Would you be able to enable that command line parameter on the test server?

davidplowman:
Right, I've done that and rebooted it. Give it a couple of minutes and cross your fingers!

chrisruk:
Thanks for doing that, looks like it ran the DRM tests :)

```
Test list file X: /home/pi/_work/picamera2/picamera2/tests/test_list.txt
Test list file DRM: /home/pi/_work/picamera2/picamera2/tests/test_list_drm.txt
Running 55 tests

Running /home/pi/_work/picamera2/picamera2/examples/overlay_drm.py... 	PASSED
Running /home/pi/_work/picamera2/picamera2/examples/preview_drm.py... 	PASSED
Running /home/pi/_work/picamera2/picamera2/examples/capture_circular_nooutput.py... 	PASSED
...
```

davidplowman:
Yes. looking good. Do you want to merge?

chrisruk:
Yeah, if you could merge that'd be cool.

PR #411, opened 2022-11-08
Update opencv_face_detect.py

edward-ardu:
Not increasing will cause opencv to fail to display.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ If you have installed _Picamera2_ previously using `pip`, then you should also u
 
 Thereafter, you can install _Picamera2_ _with_ all the GUI (_Qt_ and _OpenGL_) dependencies using
 ```
-sudo apt install -y python3-picamera2`
+sudo apt install -y python3-picamera2
 ```
 If you do _not_ want the GUI dependencies, use
 ```
--- a/examples/opencv_face_detect.py
+++ b/examples/opencv_face_detect.py
@@ -23,3 +23,4 @@ while True:
         cv2.rectangle(im, (x, y), (x + w, y + h), (0, 255, 0))
 
     cv2.imshow("Camera", im)
+    cv2.waitKey(1)


davidplowman:
Hi, thank you for this PR. Could you perhaps explain a little more why it's necessary and what happens if you don't have it? I only ask because the script works for me as it stands (and so far as I know, for most other people too), so I'm just curious to understand what might be happening. Thanks!

edward-ardu:
Because I can't upload pictures here, so I upload it to a website.

When I don't increment waitKey :

https://ibb.co/nn15bw9

I printed the data and it does exist, but opencv just shows a black window

https://ibb.co/1nTZLYQ

When I increase waitKey after

https://ibb.co/5hwVT8Q

normal work

https://ibb.co/p45NL09

I guess this won't show up on all devices, but it's definitely a possible problem.

davidplowman:
Thanks for the extra info. What kind of Pi are you using, by the way, and what kind of camera? I'm just wondering if the issue could be that the system is so busy, that it never actually manages to do the work to render the image. Does it have any effect if you use something like `time.sleep(0.01)` instead of `cv2.waitKey(1)`?

edward-ardu:
I am using a raspberry pi4 4G and the camera used is the official imx219.
https://ibb.co/Xy9s90Z

I'm using a fresh image and I haven't done any extra background work.

I followed your advice time.sleep(0.01) instead of cv2.waitKey(1), nothing worked.
https://ibb.co/VvTbJgG

edward-ardu:
I'm not an expert in opencv, so I'm not sure what's going on, this solution is what I searched in google.

I found a possible reason
https://stackoverflow.com/questions/44854699/why-does-the-cv2-imshow-does-not-render-without-cv2-waitkey

davidplowman:
Hi, thanks for the update. Honestly I'm a bit baffled by all this. It's not that I mind the change, it's just that I can't see a reason for it and I can't reproduce the problem. Here's what I know.

I too read quite a lot of stuff about OpenCV, and there's lots about `cv2.waitKey`. My interpretation of everything that I read was that it _shouldn't_ be necessary if you use `cv2.startWindowThread` first, but it's difficult to find a very clear answer on this. (And if anyone else reading this knows a definitive answer, please say!!)

I've run this example on a fresh install (Raspberry Pi OS, 64-bit full version), I've run it locally and via VNC (as you are doing), also using the imx219, but I can't make it go wrong. I also notice some strange messages on your screen shot:

* The `ERROR DelayedControls` message looks like quite a serious error, and I don't understand how that can happen.
* You also have some Qt errors further down, and I don't see these either.

That first message lists the number `0x009a090a`, and when I look that up it seems to correspond to `V4L2_CID_FOCUS_ABSOLUTE`. I am suspecting that these are not standard Raspberry Pi versions of software - perhaps you could confirm that for me. Could you say exactly what you are running? Thank you!

edward-ardu:
I restarted and burned a brand new system again, and the problem did not exist. It's amazing, I kept all the old environments, I think I'll look into this stuff more when I have time.

edward-ardu:
Anyway, thanks for your reply

davidplowman:
Thanks for the update. I'll close this for now then but do let us know if you ever find out why it behaved strangely!

PR #413, opened 2022-11-08
Global camera info

davidplowman:
These commits add a `global_camera_info` static method which should make it easier to find out what cameras are attached and which to use. There are actually 3 commits:

* The first just makes our CameraManager up front so it's always there. I was going to make it hang on to the "camera manager singleton" persistently but had to shelve that bit as there are still some issues down in libcamera.
* The second commit adds the `global_camera_info` method. I thought this was probably enough and it didn't require a separate "get the number of cameras" method.
* Finally there's just a very trivial test that uses the new method.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -186,6 +186,19 @@ class Picamera2:
             return tuning[name]
         return next(algo for algo in tuning["algorithms"] if name in algo)[name]
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


davidplowman:
@chrisruk So I think this is the last thing on my list before pulling the latest libcamera and making up a new release. Does that work for you?

chrisruk:
Sounds good, that's fine with me

PR #414, opened 2022-11-08
Small change to avoid QT message

chrisruk:
Switched to using deleteLater() rather than del on QSocketNotifier to avoid:

QSocketNotifier: Invalid socket 53 and type 'Read', disabling...

### DIFF
--- a/picamera2/previews/q_gl_picamera2.py
+++ b/picamera2/previews/q_gl_picamera2.py
@@ -119,7 +119,7 @@ class QGlPicamera2(QWidget):
     def cleanup(self):
         with self.lock:
             self.running = False
-            del self.camera_notifier
+            self.camera_notifier.deleteLater()
             eglDestroySurface(self.egl.display, self.surface)
             self.surface = None
             # We may be hanging on to a request, return it to the camera system.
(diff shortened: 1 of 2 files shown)


chrisruk:
I tested that the message had hopefully gone away by running the following up to around ~50 iterations.

```
#!/bin/bash

i=0
while [[ 1 ]]; do
	echo "At $i"
	if python3.9 close_test_multiple-1.py |& grep -q -i QSocket; then
		echo "error"
		break
	fi
	i=$((i+1))
done
```

Using this python (a derivative of close_multiple_test.py that runs a bit faster, because it's not running as many tests) -

```
#!/usr/bin/python3

import multiprocessing
import threading
import time

def run_camera(idx):
    camera = Picamera2(idx)
    camera.start_preview(Preview.QT)
    camera.start()
    time.sleep(3)
    camera.stop()
    camera.close()

if __name__ == '__main__':
    multiprocessing.set_start_method('spawn')

    print("Test two threads")
    threads = []
    for i in range(2):
        thread = threading.Thread(target=run_camera, args=(i,), daemon=True)
        thread.start()
        threads += [thread]
    for thread in threads:
        thread.join()

    print("Test camera in main process and thread")
    camera = Picamera2(0)
    camera.start_preview(Preview.QTGL)
    camera.start()
    thread = threading.Thread(target=run_camera, args=(1,), daemon=True)
    thread.start()
    thread.join()
    camera.stop()
    camera.close()
```

PR #423, opened 2022-11-11
a couple of tweaks to make run_tests more convenient

davidplowman:
@chrisruk Hi Chris, thanks for all the recent work you did extending run_tests!

Whenever I've got new release apt packages to try I always do a full clean OS install on a Pi and then run these tests. On this last occasion I found the script a bit less convenient so I've tweaked a couple of things:

* It makes the picamera2_tests dir for you if it doesn't exist (rather than failing). This seems a bit more helpful, especially where folks are running it for the first time.
* I've stopped it checking the test file names so early, meaning that you can run the script from anywhere. It still fails of course, but do you think the error (a FileNotFound from load_test_list) is clear enough?

### DIFF
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -121,20 +121,20 @@ def directoryexists(arg):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='picamera2 automated tests')
-    parser.add_argument('--dir', '-d', action='store', type=directoryexists,
+    parser.add_argument('--dir', '-d', action='store',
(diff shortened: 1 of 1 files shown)


davidplowman:
Hmm, the test machine seems to be having a bit of difficulty. Not sure what's going on there. Will check up on it on Monday.

chrisruk:
Hi David, I'm fine with this change, sounds a good idea re. creating the directory.  Will get it merged when the test machine is running again.

PR #429, opened 2022-11-16
Add some simple multi-camera tests

davidplowman:
But first there's a commit which allows tests to be "skipped", so it's safe to run all the tests whatever devices you currently have plugged in.

The second commit adds a couple of fairly trivial multi-cam tests into the test list.

### DIFF
--- a/tests/close_test_multiple.py
+++ b/tests/close_test_multiple.py
@@ -17,6 +17,11 @@ def run_camera(idx):
 
 
 if __name__ == '__main__':
+
+    if len(Picamera2.global_camera_info()) <= 1:
+        print("SKIPPED (one camera)")
+        quit()
+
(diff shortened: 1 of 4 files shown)


davidplowman:
Hmm, the only weird thing is that the "multcamera" tests are saying the were skipped.... which they're not supposed to! Let me look at that...

chrisruk:
Would you be able to add 'tests/close_test_multiple.py' to the list too? Thanks!

davidplowman:
I think it's that "Unable to set controls" error that is making the test fail. There is work happening to get rid of this error down in libcamera, but I think for now we probably just need to "escape" this one...

davidplowman:
I think order has been restored now!!

Issue #426, opened 2022-11-14
[BUG] `picamera2.capture_***()` can return frames from before the call


meawoppl:
**Please only report one bug per issue!**

**Describe the bug**
Picamera2 capture methods can return images taken before `capture_***` is called.

**To Reproduce**
Call any capture method, and inspect the timestamp vs one taken before call. It is extra obvious for long exposures.

**Expected behaviour**
Methods return captures from after the method call.

**Hardware :**
All cameras/os are impacted.

**Additional context**
Requests are internally stored/queued [here](https://github.com/raspberrypi/picamera2/blob/main/picamera2/picamera2.py#L1137). In the case of long exposures, the information can be from seconds to tens of seconds before the call is made. Highly unexpected.

This impacts encoder methods as well.

davidplowman:
Hi, thanks for reporting this. Sometimes this is a useful features but of sometimes, I agree, it might be something an application would want to avoid. A few ideas spring to mind:

* You could add extra flags to capture methods to say "don't do that" (not sure I like that too much).
* We could have a `flush_captures` method that discards the queue (it would be a bit like calling `capture_metadata`, except that the latter would wait for a frame if there wasn't one queued up).
* Or perhaps it's something we could tie to the configuration? We might make "use a queue" the default for preview/video, and perhaps "don't" for stills, for example, though we'd make make sure that you can override it.

Any thoughts?

PR #430, opened 2022-11-16
Let the camera config decide whether we hold on to the last request

davidplowman:
A flag in the configuration determines whether the last received completed request is kept or not. When it is kept, it can be returned quickly when a new capture request occurs. Otherwise capture requests will always wait for a new frame to arrive.

By default we always enable the queue so as not to change existing behaviour unless requested. This can of course be overridden when the configuration is created using the "queue" parameter.

### DIFF
--- a/picamera2/configuration.py
+++ b/picamera2/configuration.py
@@ -85,7 +85,7 @@ class StreamConfiguration(Configuration):
 
 
 class CameraConfiguration(Configuration):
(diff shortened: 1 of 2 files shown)


davidplowman:
One could go further with this, such as making the queue length fully adjustable, even while running, but I'm disinclined to do anything as dramatic as that unless there's a demonstrated need.

This change is intended to address https://github.com/raspberrypi/picamera2/issues/426.

PR #433, opened 2022-11-18
Pass all available requests to encoder

chrisruk:


picamera2/picamera2.py
```diff
@@ -1119,10 +1119,11 @@ def process_requests(self) -> None:
...
             display_request.acquire()
 
-            # Some applications may (for example) want us to draw something onto these images before
-            # encoding or copying them for an application.
-            if display_request and self.pre_callback:
-                self.pre_callback(display_request)
+            for req in requests:
+                # Some applications may (for example) want us to draw something onto these images before
+                # encoding or copying them for an application.
+                if self.pre_callback:
```
davidplowman, line 1125: Very tiny nit-pick... maybe do this check outside the loop?

picamera2/picamera2.py
```diff
@@ -1141,14 +1142,18 @@ def process_requests(self) -> None:
...
+                    continue
 
-            if self._encoder is not None:
-                stream = self.stream_map[self.encode_stream_name]
-                self._encoder.encode(stream, display_request)
+                if self.post_callback:
+                    self.post_callback(req)
+
+                if self._encoder is not None:
+                    stream = self.stream_map[self.encode_stream_name]
```
davidplowman, line 1155: Same again, maybe set `stream` outside the loop?

picamera2/picamera2.py
```diff
@@ -1141,14 +1142,18 @@ def process_requests(self) -> None:
...
                     self._future.set_exception(e)
 
-            # Some applications may want to do something to the image after they've had a change
-            # to copy it, but before it goes to the video encoder.
-            if display_request and self.post_callback:
-                self.post_callback(display_request)
+            for req in requests:
+                # Some applications may want to do something to the image after they've had a change
+                # to copy it, but before it goes to the video encoder.
+                if req not in self.completed_requests:
```
davidplowman, line 1148: I was curious about this, can it happen?

picamera2/picamera2.py
chrisruk: Good point, definitely makes sense to do it the other way round.  (Will be back in on Wednesday and will sort then)

picamera2/picamera2.py
chrisruk: Yeah, this seemed to happen with one of the opencv examples, it can take off one of the requests, via capture_buffer if I recall correctly.

davidplowman: My feeling is that even if someone "eats" the capture request, we should probably still pass it to the encoder. Otherwise if someone were to capture a still while a video is recording, that frame would disappear from the video. What do you think?

chrisruk: Yeah I did wonder that, however, if you remove the line
```
if req not in self.completed_requests:
```

When running -
python3.9 examples/opencv_face_detect_2.py 

We get
```
  File "/home/pi/Repositories/picamera2/picamera2/previews/q_gl_picamera2.py", line 362, in handle_requests
    request = self.picamera2.process_requests()
  File "/home/pi/Repositories/picamera2/picamera2/picamera2.py", line 1155, in process_requests
    self.post_callback(req)
  File "/home/pi/Repositories/picamera2/examples/opencv_face_detect_2.py", line 17, in draw_faces
    with MappedArray(request, "main") as m:
  File "/home/pi/Repositories/picamera2/picamera2/request.py", line 45, in __init__
    self.__buffer = _MappedBuffer(request, stream)
  File "/home/pi/Repositories/picamera2/picamera2/request.py", line 17, in __init__
    self.__fb = request.request.buffers[stream]
AttributeError: 'NoneType' object has no attribute 'buffers'
```

davidplowman: I think we're probably not handling our reference counts properly for this new behaviour. The CompletedRequests get created with a reference count of 1, which gets "transferred" to "being in the completed_requests list". But if a capture comes along and removes it, then it may get recycled before we run the post_callback. So we need to tweak that a bit. Maybe that initial count of 1 gets transferred to "being in the requests lists", and being in the completed_requests list counts as an extra one? Something like that...

chrisruk: Yeah I just tried bumping the refcount prior to adding to the completed requests, then decrementing after running the loop, which meant I could remove that if statement.

Just pushed changes again to this pull req.

picamera2/picamera2.py
```diff
@@ -1141,14 +1145,20 @@ def process_requests(self) -> None:
...
+                # Some applications may want to do something to the image after they've had a change
+                # to copy it, but before it goes to the video encoder.
+                if self.post_callback:
+                    self.post_callback(req)
+
+                if self._encoder is not None:
+                    self._encoder.encode(stream, req)
+
+            for req in requests:
+                req.release()
```
davidplowman, line 1161: Maybe this could be the last line of the previous loop? (Sorry for the nit-pick!)

But otherwise the reference counting looks correct to me... :)

chrisruk: Good point, just changed.

chrisruk:
Thanks :) I'm ok with it now.

PR #436, opened 2022-11-21
Replace the list of functions to execute by a "job" class

davidplowman:
There are 5 commits here but I think they leave things tidier and better. I've tried to organise the changes and commit messages so that it's reasonably clear what's happened, and the tests should be fully working after each commit. They are:

1. This is just a simple code tidy in this general area, no functional changes.
2. Replaces the list of functions that get executed by a "job". Again there is no functional change here, but wrapping up all the functions into a single "job" makes things tidier, and will help subsequent commits.
3. Here we update the `wait()` method. You now have to say what "job" you're waiting for. This is a breaking change but this commit also fixes up all the Qt examples that are affected. In particular, we have to pass the "job" through the Qt signals to the application's callback functions.
4. With all the functions now in a single "job", it's easy to extend the single job to a list, so that you can have several at once.
5. Adds a test for these changes.

### DIFF
--- a/picamera2/job.py
+++ b/picamera2/job.py
@@ -1,0 +1,63 @@
+from concurrent.futures import Future
+
+
+class Job:
+    """
+    A Job is an operation that can be delegated to the camera event loop to
+    perform, such as capturing and returning an image. Most jobs only do a single
+    thing, like copying out a numpy array, and so consist of a single function
+    that we pass to do this.
+
+    But some jobs may take several frames to complete, for example if they involve
+    mode switches, or waiting for certain controls to take effect. Here we need
+    to submit a list of functions to perform, and as each function, representing
+    a stage of the job, completes, then we will move on to the next function in
+    the list when the next frame arrives.
+
+    Jobs are normally created by the Picamera2.dispatch_functions method, though
+    most common operations have dedicated methods to do this, such as
+    Picamera2.switch_mode_and_capture_array.
+    """
+
+    def __init__(self, functions, signal_function=None):
+        self._functions = functions
+        self._future = Future()
+        self._future.set_running_or_notify_cancel()
+        if signal_function is not None:
+            self._future.add_done_callback(signal_function)
+
+        # I wonder if there is any useful information we could collect, number
+        # of frames it took for things to finish, maybe intermediate results...
+        self.calls = 0
+
+    def execute(self):
+        """Try to execute this Job. It will return True if it finishes,
+        or False if it needs to be tried again."""
+        assert self._functions, "Job already completed!"
+
+        try:
+            # Each function making up the Job returns two things: whether it's
+            # "done", in which case we pop it off the list so that the function
+            # in the list will run with the next frame (otherwise we leave it there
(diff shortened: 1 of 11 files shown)


davidplowman:
@chrisruk This should fix that little test case you sent to me the other day.

picamera2/job.py
```diff
@@ -0,0 +1,37 @@
...
+        try:
+            done, result = self._functions[0]()
+            self.calls += 1
+            if done:
+                self._functions.pop(0)
+
+                if not self._functions:
+                    self._future.set_result(result)
+
+        except Exception as e:
```
chrisruk, line 30: I assume 'Exception' can't be narrowed to a subclass? Since arbitrary functions can be passed?

chrisruk (commented):
I was a little confused about the multiple jobs list, as a single job can run multiple functions already?

davidplowman:
> I was a little confused about the multiple jobs list, as a single job can run multiple functions already?

I wonder if some more comments.explanation would help in job.py. The idea is that you can schedule a job that takes more than a single frame to complete. For example, you might mode-switch the camera, do a capture, then mode-switch back to the preview mode. Or you might wait for particular settings that you have sent to "take effect" (which might take several frames - there's work going on in libcamera on this). (Obviously, this has always been how it works, but you've been implicitly restricted to one job at a time.)

chrisruk:
Oh, cool, gotcha, that sounds handy, re. jobs that take >1 frame to complete.  Yeah if you don't mind some more comments would be good then.

Would an example for -  'you might mode-switch the camera, do a capture, then mode-switch back to the preview mode', be easy to create for another commit, if so, would be interested to see how that'd be done.

davidplowman:
> ...
> Would an example for - `you might mode-switch the camera, do a capture, then mode-switch back to the preview mode`, be easy to create for another commit, if so, would be interested to see how that'd be done. Edit: I assumed you meant the mode switching is done in the job right?

Yes, the whole thing is done as part of the job, so all the application has to do is launch it and wait (and it's very similar in a Qt app, you can launch it but now you don't block for it to happen, you wait for your signal). As regards examples... there kind of are lots already e.g. `Picamera2.switch_mode_and_capture_array`. The revised documentation hopefully makes that a bit clearer.

chrisruk:
Hi David, by revised documentation, is that in a pdf? Cheers

davidplowman:
Sorry no, I mean the comments in job.py. Though I will update the pdf manual to talk about a "job" being returned and passed to the callback when you make an asynchronous call.

chrisruk:
Ah gotcha, was just wondering if any comments around https://github.com/raspberrypi/picamera2/blob/00368d9e078e3cf50fa362e002f3e0097a6e033a/picamera2/job.py#L36 would be useful, such as how it's used by the 'wait' method in picamera

davidplowman:
Yep, fair point, I'll put something extra in there, and in the `Picamera2.wait` method too!

davidplowman:
I've improved a couple of those doc strings.

chrisruk:
Thanks! Will merge now

PR #443, opened 2022-11-25
Add locking to the V4L2 encoder to prevent encode() running after stop()

davidplowman:
This change makes it safe to call Picamera2.stop_encoder() without stopping the camera first.

When an application calls V4L2Encoder._stop() it's possible that the camera might still be in V4L2Encoder.encode(), feeding in another frame which could then crash. This just adds a lock so that we can be sure that any trailing calls to encode() won't do anything.


davidplowman:
@chrisruk Morning Chris, I was wondering if you might take a look at this one? It comes from https://forums.raspberrypi.com/viewtopic.php?t=343390 when I found that I couldn't safely call `picam2.stop_encoder()` while the camera is running. Anyway, I'm not sure if this is the best fix or if there's something better we could do... maybe you could have a think?

chrisruk:
Sure, just having a look now

davidplowman:
I guess one thing I wonder is whether this - if we go this route - should go in the base class so that all encoders would pick up the benefit?

chrisruk:
Yeah that sounds a good plant.

Just created a test case that I believe triggers this.

Can you confirm the following is what you got - 
```
Exception during process_requests()
  File "/home/pi/Repositories/picamera2/picamera2/previews/null_preview.py", line 82, in handle_request
    completed_request = picam2.process_requests()
  File "/home/pi/Repositories/picamera2/picamera2/picamera2.py", line 1159, in process_requests
    self._encoder.encode(stream, req)
  File "/home/pi/Repositories/picamera2/picamera2/encoders/v4l2_encoder.py", line 244, in encode
    fcntl.ioctl(self.vd, VIDIOC_QBUF, buf)
OSError: [Errno 22] Invalid argument
```

davidplowman:
Yes, that's exactly it!

chrisruk:
Cool thanks, maybe we should include a test case for this in the tests? If so this is what I used to trigger it -

```
#!/usr/bin/python3
import time

from picamera2.encoders import H264Encoder

picam2 = Picamera2()
video_config = picam2.create_video_configuration()
picam2.configure(video_config)
encoder = H264Encoder(10000000)

for i in range(20):
    print(f"Attempt {i}")
    picam2.start_recording(encoder, 'test.h264')
    time.sleep(0.5)
    picam2.stop_encoder()
    time.sleep(0.5)
    picam2.stop()
```

picamera2/encoders/v4l2_encoder.py
```diff
@@ -214,32 +217,35 @@ def encode(self, stream, request):
...
-        buf.length = 1
-        buf.timestamp.secs = timestamp_us // 1000000
-        buf.timestamp.usecs = timestamp_us % 1000000
-        buf.m.planes = planes
-        buf.m.planes[0].m.fd = fd
-        buf.m.planes[0].bytesused = cfg.frame_size
-        buf.m.planes[0].length = cfg.frame_size
-        fcntl.ioctl(self.vd, VIDIOC_QBUF, buf)
-        self.buf_frame.put(request)
+        with self.lock:
```
chrisruk, line 220: Just wondering, like you say if you moved the lock for 'stop' into the Encoder class.  Would we then want this self.lock initiated in the picamera class possibly before it calls 'encode'?

chrisruk: Around this call - https://github.com/raspberrypi/picamera2/blob/next/picamera2/picamera2.py#L1159

davidplowman: I expect the lock would go in the Encoder base class. Then everyone always calls a base class method (which deals with the lock), and then calls the more internal "do the work" derived method. Here we're saying that "encode" and "stop" need this behaviour, actually I wonder if "start" does too (see my next comment).

davidplowman:
Yes, definitely need a test case. Actually one other thing occurred to me. Presumably it should be possible to start/stop the encoder _without_ starting/stopping the camera. So something like this (very similar to the above):
```
from picamera2.encoders import H264Encoder
from picamera2.outputs import FileOutput
import time

picam2 = Picamera2()
encoder = H264Encoder(10000000)
picam2.start(picam2.create_video_configuration())

for i in range(10):
    print(i)
    encoder.output = FileOutput("test.h264")
    picam2.start_encoder(encoder)
    time.sleep(0.5)
    picam2.stop_encoder()
    time.sleep(0.5)

picam2.stop()
```
Now this seems to work OK, but I think the encoder's `start_` method also wants protecting with the lock, just like `stop_` and `encode_`. Otherwise there's a risk we could be half way through when a frame turns up and goes into `encode_`.

(I also felt that we could make `start_encoder` easier to use by optionally passing in the output - like `start_recording` - but that's a separate commit.)

Are you OK to continue looking at this? I could take over if you're doing other stuff.

chrisruk:
Sure, happy to continue looking at this, shall I create a new pull request?

davidplowman:
Yes, I think we probably want to replace this one, I guess I was just testing the water!!

davidplowman:
We merged an improved version of this (#444)

PR #445, opened 2022-11-25
Fix Qt problems closing preview windows with the X button

davidplowman:
WHen the X button is clicked to close a preview window, it wasn't running through the usual cleanup operations, resulting in a nasty hang when trying to quit Python. This adds some "close event" handling to perform the necessary steps.

We also need to tell the Picamera2 object that the preview disappeared spontaneously, and make sure the Qt App stays running even when the only window is closed in this way, otherwise we have no thread to create a new one.

### DIFF
--- a/picamera2/previews/qt_previews.py
+++ b/picamera2/previews/qt_previews.py
@@ -33,10 +33,26 @@ class QtPreviewBase:
 
         @QtCore.pyqtSlot()
         def createpreview(parent, cam, previewretrieveq):
+            from PyQt5.QtGui import QGuiApplication
+            # This ensures the Qt app never quits when we click the X on the window. It means
+            # that after closing the last preview in this way, the Qt app is still running and
+            # you can create new previews.
+            QGuiApplication.setQuitOnLastWindowClosed(False)
+
             qpicamera2 = parent.make_picamera2_widget(cam, width=parent.width, height=parent.height,
                                                       transform=parent.transform)
             if parent.x is not None and parent.y is not None:
                 qpicamera2.move(parent.x, parent.y)
+
+            def closeEvent(parent, preview):
+                preview.cleanup()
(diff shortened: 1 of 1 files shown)


davidplowman:
@chrisruk You probably want to look over this PR too! I can't say it's not a bit hacky at the moment...

The problems are easy enough to demonstrate, you just need to run this:
```
picam2 = Picamera2()
picam2.start_preview(Preview.QTGL)
```
Then click the X (close window) button in the corner of the preview window and finally try to quit Python. It hangs, and when you finally hit Ctrl-C it dies nastily.

For the second test, run the same thing and click the X button again. Now try and restart another preview by repeating `picam2.start_preview(Preview.QTGL)`. (All this works fine if you use `picam2.stop_preview()` instead of clicking the X button.)

The first problem I fixed by adding a closeEvent method to do some cleaning up. Though I don't think I've done it in a very "Qt" way, which would probably want a base class that implements the method conventionally, and which we then derive from.

For the second problem, I found myself wanting the Qt "app" not to quit when you close its only window. Not sure if there's a better way, disabling the "close" button altogether might not be a bad idea either. Anyway, it's probably one to sleep on and think about next week!

chrisruk:
Hi David, sure will take a look at this next week.  Have a good weekend!

chrisruk:
Looking at this, this morning, sorry for the delay

davidplowman:
No probs, actually I'm trying to do a "better" version of this. It seems more complicated, but I hope leaves things in a slightly tidier state. I'll post it in a bit.

davidplowman:
@chrisruk Hi again - don't know if you spotted it, I just posted a very heavily revised version of this (link above as a "mention"). I think that one is probably better... although there's a lot more in it... maybe let me know what you think. Thanks!!

PR #447, opened 2022-11-25
Make start_encoder parameters more like start_recording

davidplowman:
Both now take an encoder and an output, making them a bit more consistent and it's easier to remember how to use them.

It does actually break just a few existing usages of the API, but they're easily updated (as has been done as part of this commit).


chrisruk:
Looks a good idea! will merge now

PR #452, opened 2022-11-28
Clean up imports and lack of typing.

meawoppl:
Remove a confusing star import and cleanup the typing on `encoder.py`

Side question for @davidplowman:  Is there a reason that the `_` prefix is being used on methods on this class that are clearly not private? AFAICT, everything with an underscore prefix is called outside of this file...

### DIFF
--- a/picamera2/encoders/encoder.py
+++ b/picamera2/encoders/encoder.py
@@ -2,11 +2,12 @@
 
 import threading
 from enum import Enum
+from typing import Optional
 
-from v4l2 import *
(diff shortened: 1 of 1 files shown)


davidplowman:
@chrisruk Hi Chris, are you happy to merge this one? I think the style checker complaint is one that we can ignore.

As regards the use of `_`, I think many of them have actually been resolved in this file. I'm always happy to review them, however, I know there are many places where they are/aren't used appropriately for all kind of reasons, and we haven't got round to tidying them. Thanks!

chrisruk:
Thanks! Just noticed a small typo in the signature

meawoppl:
@chrisruk and @davidplowman

PR #455, opened 2022-11-28
Fix Goofy `import selectors` placement.

meawoppl:
Not sure why these are in the places that they are right now. Imports are cached anyway...


naushir:
@meawoppl you are simply wasting everyone's time with actions like this!  If you want to seriously contribute to picamera2, you need to follow our guidelines.

Any more like this, and we will have you banned from this project.

PR #458, opened 2022-11-29
Make flake8 more lenient

chrisruk:
This now makes the github workflow make use of the flake8 exit code, so flake8 will also be used in addition to the style checker in order to lint code.


chrisruk:
@davidplowman I've made it so that with this pull request flake8 will be used to lint code along with the style checker.  I've added some global flake8 exceptions in addition to some exceptions for individual files.

```
ignore = D10, D400, D401, E800, Q000, S311, PLW, PLC, PLR, PLE
```

* D10 - enforce docstrings
* D400 - First line should end with a period
* D401 - First line should be in imperative mood
* E800 - found commented out code
* Q000 - Remove bad quotes

davidplowman:
Hi Chris, mostly this looks really good, and I assume it's going to keep us much more on our toes which is great.

Just the odd question - are there any places where we think flake8 and the existing style_checker disagree? I assume you've either "escaped" these already, or we could do so as needed.

I'll have a little look through all the changes, if there are any places where we don't like what it's done then presumably we can escape those?

Finally, and slightly longer term, is there a point in future where it becomes worth using just this and abandoning the existing style checker?

picamera2/previews/drm_preview.py
```diff
@@ -206,8 +206,8 @@ def render_drm(self, picam2, completed_request):
         if self.overlay_fb is not None:
             width, height = self.overlay_fb.width, self.overlay_fb.height
             self.crtc.set_plane(self.overlay_plane, self.overlay_fb, x, y, w, h, 0, 0, width, height)
-        overlay_old_fb = None  # The new one has been sent so it's safe to let this go now
```
davidplowman, line 209: Just curious as to what it objected to here...?

picamera2/request.py
```diff
@@ -168,8 +166,10 @@ def save_dng(self, filename, name="raw"):
 
 
 class Helpers:
-    """This class implements functionality required by the CompletedRequest methods, but
-    in such a way that it can be usefully accessed even without a CompletedRequest object."""
+    """This class implements functionality required by the CompletedRequest methods.
+
+    In such a way that it can be usefully accessed even without a CompletedRequest object.
```
davidplowman, line 171: Did flake8 reword this and change the grammar? It's obviously magic (though I'm not totally sure about the result)!

picamera2/converters.py
```diff
@@ -1,8 +1,8 @@
 import numpy as np
 
-YUV2RGB_JPEG      = np.array([[1.0,   1.0,   1.0  ], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
+YUV2RGB_JPEG = np.array([[1.0, 1.0, 1.0], [0.0, -0.344, 1.772], [1.402, -0.714, 0.0]])
```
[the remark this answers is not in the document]
chrisruk: Good question, the existing style checker didn't like the spacing too.  With flake8 though, I can add '# noqa' to make it allow this.  I could turn off the existing style checker now if you want, as I think flake8 should be stricter than the existing style checker, as it can catch things like imports that aren't needed etc. as far as I understand?

picamera2/request.py
chrisruk: Sorry I had had to do that manually, so there's a summary line. I agree that needs improving how I've done that.

picamera2/picamera2.py
```diff
@@ -187,9 +190,9 @@ def find_tuning_algo(tuning: dict, name: str) -> dict:
 
     @staticmethod
     def global_camera_info() -> list:
-        """
-        Return Id string and Model name for all attached cameras, one dict per camera,
-        and ordered correctly by camera number. Also return the location and rotation
+        """Return Id string and Model name for all attached cameras, one dict per camera.
+
+        Ordered correctly by camera number. Also return the location and rotation
```
davidplowman, line 195: The original seems to read more naturally to me...

chrisruk:
With flake8 I can make it ignore specific error codes, or you can add ```# noqa``` after a line, to make it allow that line.

The existing style checker seemed ok about most things, as I think flake8 is stricter than it.

picamera2/picamera2.py
chrisruk: I could look at reverting these comment changes if you'd like and add an exception, so we don't have to have
a summary line?

picamera2/previews/drm_preview.py
chrisruk: I think that's because ```overlay_old_fb``` isn't read by anything.  But I see that you're using it to hold onto an object temporarily.

picamera2/picamera2.py
davidplowman: Ah, I guess I hadn't understood what's happening. It expects a "summary line" as the first one in the comment, and then a blank line and then the rest? Maybe we need to edit these just a bit so they don't become a bit ungrammatical, but I'm happy to adopt this going forward.

davidplowman:
I think I'd be happy to do that. Couple of things related to that (to save me googling - sorry!), how would I run this locally on my Pi to stop me pushing loads of rubbish, and can I put something into my `.git/hooks/pre-commit` so that I can't forget?! Thanks!

chrisruk:
Cool, so if you do --
```
pip3 install -r requirements-test.txt 
```
That should install flake8 and its various plugins for you.

Then you should be able to run --

```
flake8 . --statistics --count
```

That's a good question re. git hook, let me just try that and get back to you.

chrisruk:
Sorry about this, had to update the pre-commit yaml file again, as it seems the flake8 pre-commit hook runs in its own venv, so I had to add the plugins I added to the pre-commit yaml

davidplowman:
Cool, I'm thinking we should merge this pretty soon, and to some extent if it's done anything we don't like we can put them back again after.

I'd just like to give it a few days because I made up an updated release yesterday but haven't quite got round to testing the final apt packages. Maybe once that's done we can go for it.

chrisruk:
@davidplowman Good morning, just wondering should I remove the style checker workflow, and revert the matrix whitespace columns in this pull req.

davidplowman:
Hi Chris, yes I think so, let's try that. Then I'll pull down the branch and give it a try, make sure I can still make commits and stuff like that!

davidplowman:
Thanks! Just one last request before I try it...

Could you update the section at the end of the  README.md to give updated instructions for how to install flake8 and how to use it as a git hook? Thanks!

chrisruk:
Good point, will do

chrisruk:
I've just done that.  Something I did just think, the flake8 in the github action, will lint all files, however I guess since it only takes around 3mins in total to run, this is ok, since the smoke tests take a lot longer.

(The flake8 in the pre-commit hook runs much faster though)

chrisruk:
Hi David, yep, am happy for this to be merged.

PR #461, opened 2022-12-02
Fix Qt problems closing preview windows with the X button (version 2)

davidplowman:
This is an alternative version of https://github.com/raspberrypi/picamera2/pull/445.

In some respects I've changed a lot more in this version, though hopefully things are left in a slightly better state. There are 6 commits:

1. The first adds attach_preview and detach_preview methods that the Picamera2 object uses.
2. Responsibility for calling these is moved to the previews, not the Picamera2 object. That's because the Qt previews can be closed "spontaneously" and will therefore need this code path - so I've changed it always to work like this.
3. Now we can add closeEvent handling so that the cleanup happens when a Qt window disappears spontaneously.
4. This fixes the problem where you couldn't create a new preview after one closed "spontaneously".
5. Mostly cosmetic, but I removed the "asynchronous" property because it's only used once. The commit message documents what happens when previews are started/stopped, which is now better defined.
6. Add a test for "spontaneous" closing of Qt previews.

TBH, I find the whole Qt thing, with the difficulties caused by running everything in one thread, which isn't the main thread, with multiple windows, really quite nasty. On the other hand, at least the processes involved in starting and stopping them is clearer now so that, given how much we've learnt, there's always the possibility that we could do a nicer version in the future at some point.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -536,8 +536,8 @@ class Picamera2:
             # Assume it's already a preview object.
             pass
 
+        # The preview windows call the attach_preview method.
         preview.start(self)
-        self.attach_preview(preview)
 
     def detach_preview(self) -> None:
         self._preview = None
@@ -552,8 +552,8 @@ class Picamera2:
             raise RuntimeError("No preview specified.")
 
         try:
+            # The preview windows call the detach_preview method.
             self._preview.stop()
-            self.detach_preview()
         except Exception:
             raise RuntimeError("Unable to stop preview.")
 
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -494,6 +494,10 @@ class Picamera2:
                 self.sensor_modes_.append(cam_mode)
         return self.sensor_modes_
 
+    def attach_preview(self, preview) -> None:
+        self._preview = preview
+        self.have_event_loop = True
+
     def start_preview(self, preview=False, **kwargs) -> None:
         """
         Start the given preview which drives the camera processing.
@@ -533,8 +537,11 @@ class Picamera2:
             pass
 
         preview.start(self)
-        self._preview = preview
-        self.have_event_loop = True
+        self.attach_preview(preview)
+
+    def detach_preview(self) -> None:
+        self._preview = None
+        self.have_event_loop = False
 
     def stop_preview(self) -> None:
         """Stop preview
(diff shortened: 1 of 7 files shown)


picamera2/previews/qt_previews.py
```diff
@@ -105,8 +105,6 @@ def stop(self):
             retq = Queue()
             QtPreviewBase.previewcreateq.put((Command.DELETE, retq, (self, self.qpicamera2)))
             retq.get()
-            del self.qpicamera2
```
chrisruk, line 108: Just wondering why this part was removed? I thought this was necessary to remove 
the preview when stop was called.

chrisruk (commented):
Looks good!

picamera2/picamera2.py
```diff
@@ -491,6 +491,10 @@ def sensor_modes(self) -> list:
                 self.sensor_modes_.append(cam_mode)
         return self.sensor_modes_
 
+    def attach_preview(self, preview) -> None:
```
chrisruk, line 483: Just wondering, what happens if we have >1 preview window?

picamera2/previews/qt_previews.py
davidplowman: Yes, good question. I wish I had a convincing answer...!

I decided to make it the widget's job to do all the cleaning up, so once you get here after the `Command.DELETE` has happened, `self.qpicamera2` is already `None`. So I would have to store `self.qpicamera2` beforehand, and then delete it after. Or maybe the `qpicamera2`'s `cleanup` method would delete itself? It all felt a bit weird.

So anyway, I wondered whether I really needed it if I have dropped all the handles to it... and it _seems_ like I don't. At least, all the tests pass, and we did go to some effort to test opening/closing things. But I'm well aware that this isn't conclusive. All in all I'm a bit nervous of all this Qt stuff, so more testing is certainly warranted...

picamera2/previews/qt_previews.py
```diff
@@ -53,6 +53,12 @@ def __init__(self, previewcreateq, app):
                 self.app = app
 
             def run(self):
+                from PyQt5.QtGui import QGuiApplication
```
chrisruk, line 56: Looks good!

picamera2/picamera2.py
davidplowman: I don't think that can happen. If we have several preview windows, it's because we have several cameras, and several Picamera2 objects with one each. So long as we're checking that there isn't already a preview when we make a new one, then I think we're OK?

chrisruk: Oh gotcha, sorry forgot each preview window is for a different picamera instance.

davidplowman:
I'll just make this PR flake8-friendly and then re-push it.

chrisruk:
Thanks, looks good, is it ok for me to close the old pull req. and merge this one?

PR #481, opened 2022-12-15
Move Examples into Tests Folder

meawoppl:


examples/mjpeg_server.py
```diff
@@ -1,96 +0,0 @@
-#!/usr/bin/python3
```
meawoppl, line 1: This file is duplicated.

PR #510, opened 2023-01-10
Work around OpenGL error on 64-bit OS Lite builds

davidplowman:
The 64-bit OS Lite build encounters an OpenGL error when trying to start a remote preview window. This just works around the problem, letting the non-OpenGL preview work as expected.

The 32-bit OS Lite works, so the root cause of the error is not really understood.

### DIFF
--- a/picamera2/previews/qt.py
+++ b/picamera2/previews/qt.py
@@ -1,2 +1,15 @@
-from .q_gl_picamera2 import EglState, QGlPicamera2
+# Installing Qt and OpenGL on the 64-bit Lite OS, and trying to run
+# a remote preview window, causes an error here (the 32-bit Lite OS is OK).
+# It may be something to do with more recent versions of python3-opengl?
+# Anyway, if we carry on regardless at least the non-OpenGL preview works,
+# which is in any case what is required for remote preview windows.
+from logging import getLogger
+
 from .q_picamera2 import QPicamera2
+
+_log = getLogger(__name__)
+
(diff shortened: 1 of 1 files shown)


davidplowman:
@chrisruk Hi Chris, don't know what to make of these flake8 errors... does it mean anything to you?

chrisruk:
@davidplowman Hi, it looked like one of the dependencies of pylint (astroid) has an issue with the latest version, I've just created a pull request for using an earlier version of this

davidplowman:
Ah, interesting. Well, please just merge that if it seems to resolve the problem. Thanks!!

picamera2/previews/qt.py
```diff
@@ -1,2 +1,11 @@
-from .q_gl_picamera2 import EglState, QGlPicamera2
+# Installing Qt and OpenGL on the 64-bit Lite OS, and trying to run
+# a remote preview window, causes an error here (the 32-bit Lite OS is OK).
+# It may be something to do with more recent versions of python3-opengl?
+# Anyway, if we carry on regardless at least the non-OpenGL preview works,
+# which is in any case what is required for remote preview windows.
+try:
+    from .q_gl_picamera2 import EglState, QGlPicamera2
+except Exception:
+    print("Warning: OpenGL will not be available")
```
chrisruk, line 9: Just wondering, should we use the logger here, rather than print, or is that not possible here?

PR #512, opened 2023-01-10
Imx708 test fixes

davidplowman:
A couple of tests were failing with the imx708 because it makes images wider than the GPU can handle.


davidplowman:
Ah, that's failed again. I guess I need to do a rebase and then re-push.

PR #515, opened 2023-01-12
Convert values in the camera_properties to Python-friendly types

davidplowman:
For example, libcamera.Rectangle objects are converted to tuples of 4 numbers, as we do for camera_controls. This just makes things more consistent.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -328,7 +328,9 @@ class Picamera2:
 
     @property
     def camera_controls(self) -> dict:
-        return {k: (v[1].min, v[1].max, v[1].default) for k, v in self.camera_ctrl_info.items()}
+        return {k: (convert_from_libcamera_type(v[1].min),
+                    convert_from_libcamera_type(v[1].max),
+                    convert_from_libcamera_type(v[1].default)) for k, v in self.camera_ctrl_info.items()}
 
     @property
     def title_fields(self):


picamera2/picamera2.py
```diff
@@ -328,7 +328,10 @@ def camera_properties(self) -> dict:
 
     @property
     def camera_controls(self) -> dict:
-        return {k: (v[1].min, v[1].max, v[1].default) for k, v in self.camera_ctrl_info.items()}
+        def convert_values(value):
+            values = (value[1].min, value[1].max, value[1].default)
+            return tuple(map(convert_from_libcamera_type, values))
```
chrisruk, line 333: I'm a little bit confused by this, you're using map, to call convert_from_libcamera_type on a tuple of values, 
won't convert_from_libcamera_type always return the same tuple as is passed to it in this case?

Since it looks like at tuple wouldn't be caught by any of the 'if' statements in - https://github.com/raspberrypi/picamera2/blob/next/picamera2/utils.py#L5

davidplowman: Hi Chris, thanks for querying this. Perhaps I was trying too hard not to write the slightly "obvious but tedious" version:
```
    def camera_controls(self) -> dict:
        return {k: (convert_from_libcamera_type(v[1].min),
                    convert_from_libcamera_type(v[1].max),
                    convert_from_libcamera_type(v[1].default)) for k, v in self.camera_ctrl_info.items()}
```
But honestly it looks fine so maybe it would be better. The particular case I have is where each of those (`v[1].min` for example) is a `libcamera.Rectangle` and elsewhere we're converting those into tuples of 4 numbers `(x, y, width, height)`.

PS. Did you get your imx708s?

chrisruk: Oh sorry, I see how it works now, you're saying v[1].min is a Rectangle type, and the map is converting it one element by one?

Yes I got the imx708s, thanks a lot!

davidplowman: True, but the mere fact that it was hard to understand makes it band, so let me replace it by the "obvious" version!

PR #517, opened 2023-01-12
Add more type hints

CrenoEno:
Greetings!

As type hints are pretty nice to have IMO, I added some type hints to various functions and methods.
This should make it easier for the user to use the `picamera` functionality as well as for the maintainers to refactor existing code.

As the project is on Python 3.9+, the built-in types like `dict`, `list` and `tuple` can be used directly instead of using the deprecated `typing` types like [typing.Dict](https://docs.python.org/3/library/typing.html#typing.Dict), [typing.List](https://docs.python.org/3/library/typing.html#typing.List) and [typing.Tuple](https://docs.python.org/3/library/typing.html#typing.Tuple).

Signed-off-by: Creno Eno <creno.eno@protonmail.com>

### DIFF
--- a/picamera2/configuration.py
+++ b/picamera2/configuration.py
@@ -1,8 +1,10 @@
+from typing import Any, Union
+
 from .controls import Controls
 
 
 class Configuration:
-    def __init__(self, d={}):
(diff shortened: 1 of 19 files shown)


davidplowman:
Seems like a problem fetching ubuntu updates? Probably a case of "try again later"...

PR #523, opened 2023-01-16
Add MJPEG server example using h/w MJPEG encoder

davidplowman:
Otherwise identical to mjpeg_server.py.


CrenoEno:
Note: [flake8 bugbear](https://github.com/PyCQA/flake8-bugbear#list-of-warnings) has released its `23.1.14` version that includes the new `B028` error.

davidplowman:
@CrenoEno Thanks for the info. Perhaps I'll just go and fix those up first!

PR #529, opened 2023-01-18
Allow H264 headers to specify a framerate

davidplowman:
This is merely a number in the H264 SPS headers - it may or may not actually be true. The codec has no idea when frames will actually be delivered to it.

It's optional whether to supply this number, and in most cases, unless you specifically need a value in the headers, we would recommend omitting it and driving the camera however you like.

### DIFF
--- a/picamera2/encoders/h264_encoder.py
+++ b/picamera2/encoders/h264_encoder.py
@@ -12,7 +12,7 @@ from picamera2.encoders.v4l2_encoder import V4L2Encoder
 class H264Encoder(V4L2Encoder):
     """Uses functionality from V4L2Encoder"""
 
-    def __init__(self, bitrate=None, repeat=True, iperiod=None):
+    def __init__(self, bitrate=None, repeat=True, iperiod=None, framerate=None, enable_sps_framerate=False):
         """H264 Encoder
 
         :param bitrate: Bitrate, default None
@@ -21,12 +21,18 @@ class H264Encoder(V4L2Encoder):
         :type repeat: bool, optional
         :param iperiod: Iperiod, defaults to None
         :type iperiod: int, optional
+        :param framerate: record a framerate in the stream (whether true or not)
+        :type framerate: float, optional
         """
         super().__init__(bitrate, V4L2_PIX_FMT_H264)
         if iperiod is not None:
             self._controls += [(V4L2_CID_MPEG_VIDEO_H264_I_PERIOD, iperiod)]
         if repeat:
             self._controls += [(V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER, 1)]
+        # The framerate can be reported in the sequence headers if enable_sps_framerate is set,
+        # but there's no guarantee that frames will be delivered to the codec at that rate!
+        self._framerate = framerate
+        self._enable_framerate = enable_sps_framerate
 
     def _setup(self, quality):
         if self._requested_bitrate is None:
--- a/picamera2/encoders/h264_encoder.py
+++ b/picamera2/encoders/h264_encoder.py
@@ -3,7 +3,10 @@
 from math import sqrt
 
 from v4l2 import (V4L2_CID_MPEG_VIDEO_H264_I_PERIOD,
-                  V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER, V4L2_PIX_FMT_H264)
+                  V4L2_CID_MPEG_VIDEO_H264_LEVEL,
+                  V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER,
+                  V4L2_MPEG_VIDEO_H264_LEVEL_4_1,
+                  V4L2_MPEG_VIDEO_H264_LEVEL_4_2, V4L2_PIX_FMT_H264)
 
 from picamera2.encoders import Quality
 from picamera2.encoders.v4l2_encoder import V4L2Encoder
@@ -33,6 +36,23 @@ class H264Encoder(V4L2Encoder):
         # but there's no guarantee that frames will be delivered to the codec at that rate!
         self._framerate = framerate
         self._enable_framerate = enable_sps_framerate
+
+    def _start(self):
+        codec_level = 40
+        # We may need to up the codec level to 4.2 if we have a guidance framerate and the
+        # required macroblocks per second is too high.
+        if self._framerate is not None:
+            mbs_per_sec = ((self._width + 15) // 16) * ((self._height + 15) // 16) * self._framerate
+            if mbs_per_sec > 245760:
+                self._controls += [(V4L2_CID_MPEG_VIDEO_H264_LEVEL, V4L2_MPEG_VIDEO_H264_LEVEL_4_2)]
+                codec_level = 42
+
+        # If the bitrate is > 10Mbps then the level must be at least 4.1
+        if self._bitrate > 10000000 and codec_level == 40:
+            self._controls += [(V4L2_CID_MPEG_VIDEO_H264_LEVEL, V4L2_MPEG_VIDEO_H264_LEVEL_4_1)]
+            codec_level = 41
+
+        super()._start()
 
     def _setup(self, quality):
         if self._requested_bitrate is None:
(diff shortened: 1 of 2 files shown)


RolandasRazma:
getting assert
```
  File "picamera2.py", line 1506, in start_recording
    self.start_encoder(encoder, output, pts=pts, quality=quality)
  File "picamera2.py", line 1465, in start_encoder
    self.encoder.start()
  File "encoders/encoder.py", line 187, in start
    self._start()
  File "encoders/v4l2_encoder.py", line 135, in _start
    fcntl.ioctl(self.vd, VIDIOC_STREAMON, typev)
ProcessLookupError: [Errno 3] No such process
```

davidplowman:
I think I got that when I asked it for high framerates. Could you post a minimal code snippet that goes wrong?

RolandasRazma:
hmm, you are absolutely right, 30 is highest I can specify, but I can record at higher than that...

davidplowman:
Though I don't know for sure, I'm guessing that the codec needs to be set to the officially correct "profile" and "level". I think the V4L2 driver handles those, but it looks to me that it may be the Python to V4L2 layer that isn't exporting all that stuff.

It looks like it defaults to level 4.0, which is capped at 245760 macroblocks/second. For your resolution, the limit there would work out at 31fps.  To get 60fps, it will need level 4.2. I'll have to go and look into that.

RolandasRazma:
yes, combination of this PR and level change does give requested FPS using `ffmpeg -i`. Saying that, it doesn't exactly work like picamera1: picamera1 would gives 39.93 for 10sec of recording at 40fps (expected - frame drops) where with proposed changes it does say exactly what you specify - for example 40.0fps. 
It's is definitely better for my purposes than no fps in headers, but I would hesitate to merge this PR as is. Maybe as you suggested it is better to allow application to set `V4L2_CAP_TIMEPERFRAME` if it's not possible to set correct one for library...

The profile setting does look valid change as a separate PR

davidplowman:
Hi, glad that seems to be working. I didn't really understand the issue about the 40fps vs 39.93fps. As far as I know, the framerate gets set in the headers right at the start and never changes, so I can't see where that number is coming from. I'm pretty sure it's nothing to do with frame drops, perhaps there's some strange rounding going on somewhere?

davidplowman:
Note that this PR now requires the latest version of python-v4l2.

@chrisruk I couldn't persuade flake8 to like the order of the imports... any ideas?

chrisruk:
@davidplowman Hi, just wondering, did you try isort on it?

davidplowman:
> @davidplowman Hi, just wondering, did you try isort on it?

Ah, I hadn't realised I can run isort for myself to "fix" my file. Thanks! Bizarrely it seems to have moved the final two symbol imports from v4l2 onto one line... I can't say I'm not a bit baffled why this is "better". But if it works for flake8, then it works for me!

davidplowman:
@RolandasRazma Are you able to create one of these short 39.93fps h.264 bitstreams so that I could download it and have a look what's in there? Also, what camera are you using?

I'm still struggling with the fact that (AFAIK) there is _no_ timing information in an h.264 bitstream except that one time-per-frame number which gets set when the codec starts. Another theory may be that when you ask the camera for 40fps you don't quite get it because the camera has to quantise everything to whole pixel lines, maybe that's why you get a different number?

It's true that the legacy code would create variable framerate streams by default, but this was changed because certain applications (notably gstreamer, I believe) could not handle them. So now they always claim 30fps unless you say otherwise. (Even without these patches, I'm seeing that the headers have the "fixed_frame_rate" flag set to 1 so these do not appear to be variable framerate streams.)

RolandasRazma:
hmm, maybe I'm doing something wrong. Without changes picamera2 `ffmpeg -I video.h264 -vf vfrdet -an -f null -` give me `VFR:0.797954 (312/79) min: 39999 max: 40000 avg: 39999` with changes `VFR:0.000000 (0/391)` for same camera/settings

In my case it make no difference, but have nagging feeling that someone else will stumble upon this

davidplowman:
> hmm, maybe I'm doing something wrong. Without changes picamera1/picamera2 `ffmpeg -I video.h264 -vf vfrdet -an -f null -` give me `VFR:0.797954 (312/79) min: 39999 max: 40000 avg: 39999` with changes `VFR:0.000000 (0/391)` for same camera/settings
> 
> Without changes FPS reported by ffmpeg as 30 on picamera2 (39.93 on picamera1 with sps_timing turned on) With changes as 40 Requested FPS: 40
> 
> In my case it make no difference, but have nagging feeling that someone else will stumble upon this
> 
> Can try to make short videos, in all permutations. I'm using imx219 (camera v2)

I was trying to find out what that `vfrdet` does. There's actually very little code there, it's just looking at the pts (timestamp) values, which of course begs the question where they come frame.

I tried `ffprobe test.h264 -show_entries frame=pkt_pts_time -of csv=p=0` but they're all reported as N/A. So I have no idea what ffmpeg is doing. Is it trying to reverse engineer some supposed duration based on bitrate? Or maybe ffmpeg doesn't really handle the case of frames not having timestamps (I've seen this sort of behaviour elsewhere), so slightly unpredictable things happen?

If you could post a file where ffprobe reports 39.93 that would be interesting. For the rest, I'm not really seeing anything "unexpected" in the h.264 bitstreams, so it's a question of finding out what ffmpeg is really telling us.

RolandasRazma:
```python
import picamera
import time

camera = picamera.PiCamera()
camera.resolution = (1280, 720)
camera.framerate = 90
camera.start_recording('picamera1.h264', format='h264', level='4.2', sps_timing=True)
camera.wait_recording(1)
time.sleep(1)
camera.wait_recording(1)
camera.stop_recording()
```

[picamera1.h264.zip](https://github.com/raspberrypi/picamera2/files/10472946/picamera1.h264.zip)

```bash
ffmpeg -i picamera1.h264 
```
87.81 fps

```bash
ffmpeg -i picamera1.h264 -vf vfrdet -an -f null -
```
VFR:0.397129 (83/126) min: 13333 max: 48000 avg: 13333

```bash
ffprobe picamera1.h264 -show_entries frame=pkt_pts_time -of csv=p=0
```
Stream #0:0: Video: h264 (High), yuv420p(progressive), 1280x720, 87.81 fps, 90 tbr, 1200k tbn, 180 tbc
N/A
N/A
...

davidplowman:
Thanks for posting that file. Looking at that h264 bitstream, all I see is that it records a framerate of exactly 90fps, and there's no material difference between the h264 headers in any of the files that I can detect, they all seem "fair enough" to me.

So it remains all a bit of a mystery exactly what ffmpeg is telling us, and I don't think I believe those marginally different values - I think the "39.93fps" stream should be played at 40fps, and the "87.81fps" stream is in reality 90fps. I can certainly believe there are in general difficulties with raw h264 bitstreams because they don't have "proper" timestamps, having seen such things elsewhere. In summary, therefore, I'm somewhat inclined to accept this PR until such time as we can understand what, if anything, is actually wrong (at which point I'm very happy to come back to it). What do you think?

RolandasRazma:
for sure this is best we can get to, but you will have questions about this... 
Maybe naming could be more distinct from FPS like "sps_framerate" (similar to enable_sps_timing") or even "timeperframe" to copy v4l2.

picamera2/encoders/h264_encoder.py
```diff
@@ -34,6 +37,23 @@ def __init__(self, bitrate=None, repeat=True, iperiod=None, framerate=None, enab
         self._framerate = framerate
         self._enable_framerate = enable_sps_framerate
 
+    def _start(self):
+        codec_level = 40
+        # We may need to up the codec level to 4.2 if we have a guidance framerate and the
+        # required macroblacks per second is too high.
```
chrisruk, line 43: small type, assume macroblocks?

davidplowman:
Ah yes, whoops! Let me fix that quickly...

RolandasRazma:
what is release process with apt (it is now recommended)?

davidplowman:
I am unfortunately out of the office for a couple of weeks, but I'll try and do a release as soon as I can after that. If you needed something sooner could you just check out the "next" branch and use that directly for now (just remember to delete it again once the release shows up).

RolandasRazma:
no rush, just wondering how release process work in general

PR #530, opened 2023-01-19
Circular buffer copy and clear function

chrisruk:


RolandasRazma:
IMO copy and clear should be separate actions, maybe `def copyandclear(self, clear: Bool=False)` ?

RolandasRazma:
don't you need to reset `self._firstframe` in clear case?

davidplowman:
Hi Chris, could you remind me what the interface is for the circular buffer output? That is how you start it running, and then how you start it recording to a file. Is there a way to stop it recording to a file without stopping the circular buffer (i.e. so that you could start recording again later)?

I'm sort of wondering vaguely if the answer here shouldn't involve recording the circular buffer to a memory buffer rather than to a file? Or something. Sorry if I'm being a bit slow, I'm just trying to understand what we really want to be doing here. Thanks!

chrisruk:
Good question, I've been re-familiarising myself with this code.

The start method comes from the Output class, and simply sets recording to True.  Frames are passed to CircularOutput via outputframe method calls, and are always appended to the internal circular buffer.

If you call start_recording, via picamera, that will call .start() on the circular output which if you have a file defined for the circular buffer, and outputtofile is True, will start writing frames after and including the first keyframe.

Since frames are always written to the internal circular buffer, calling .stop() on the CircularOutput should still append frames to the internal buffer.

In https://github.com/raspberrypi/picamera2/pull/530/commits/9ee87ece7ece0cd48ab03fa4c6f7b0e30580acbd I made use of a CircularOutput, without specifying a file.  And then dumping the frames to a file myself in the example.

@RolandasRazma That's a good question re. resetting self._firstframe on clearing the buffer, however this is just used by the FileOutput, so I'm not sure it does need to be reset, as if frames have already been written to a file, they would have started with a keyframe anyway.  @davidplowman any thoughts on this?

PR #539, opened 2023-01-25
QPicamera2: Fix QImage alignment; don't abort if cv2 unavailable

njhollinghurst:
QImage requires image scanline data to be 32-bit aligned in memory; for RGB888 it means width must be a multiple of 4 bytes. Rather than propagate the requirement to higher-level code, just trim the image.

cv2.cvtColor is needed to render YUV420 and YUYV, but applications typically don't need this for viewfinder, only post-capture display. Noisily fail to display YUV, but don't raise an exception.

Signed-off-by: Nick Hollinghurst <nick.hollinghurst@raspberrypi.com>

### DIFF
--- a/picamera2/previews/q_picamera2.py
+++ b/picamera2/previews/q_picamera2.py
@@ -1,3 +1,5 @@
+import logging
+
 import numpy as np
 from libcamera import Transform
 from PyQt5.QtCore import (QRect, QRectF, QSize, QSocketNotifier, Qt,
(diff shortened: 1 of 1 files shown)


davidplowman:
I find the complaints about imports rather hard to decipher. I normally just end up running
```
python -m isort <filename>
```
and the look what it did!

njhollinghurst:
I'd better fix that and re-push.

By the way I've just noticed that libcamera-apps also creates a QImage that could be mis-aligned! (and am struggling to find definitive documentation about non-multiple-of-4-wide QImages: is the stride silently rounded up to the next multiple of 4 bytes, or are they just not allowed?)

PR #570, opened 2023-02-15
Update to work with the new libcamera colour space scheme

davidplowman:
The raspberrypi/libcamera next branch has been updated to adopt the latest colour space implementation in libcamera. The update here is required to make Picamera2 work with it again. Note that the Picamera2 next branch probably won't work with the libcamera main branch any more, until we do another release.

There's also a test here which tests every format and colour space combination to exhaustion, so hopefully we'll notice quickly if colour spaces get broken again!

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -19,6 +19,7 @@ from libcamera import controls
 from PIL import Image
 
 import picamera2.formats as formats
+import picamera2.utils as utils
 from picamera2.encoders import Encoder, H264Encoder, MJPEGEncoder, Quality
 from picamera2.outputs import FfmpegOutput, FileOutput
 from picamera2.previews import DrmPreview, NullPreview, QtGlPreview, QtPreview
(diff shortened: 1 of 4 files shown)


davidplowman:
Ah, it's failing because I haven't updated libcamera on the test machine. Better do that.

davidplowman:
@chrisruk This is the change that will make the next branch work again with the next branch of the raspberrypi/libcamera repository. Probably worth looking at sooner rather than later because no other PR will ever pass until we merge it!

chrisruk:
Just been looking through this, looks good to me, will merge this in a tick, just wondering I assume to use 'next' after this is merged, I'll need to compile libcamera manually?

davidplowman:
Yes, you would have to make sure you've got the raspberrypi/libcamera next branch checked out, and build it manually. Hopefully it won't be too long before we can make an updated release and packages.

chrisruk:
I just noticed after merging some of the smoke tests seem to have failed, not really sure why.  I'm just trying to build libcamera now and will run them locally.

davidplowman:
You'll definitely get some failures if you're running the updated Picamera2 "next" but without the updated libcamera. But ping me if they persist!

PR #579, opened 2023-02-20
Allow constant quantiser encoding in h.264

davidplowman:
Just requires a small tweak to set the min/max QP parameters, and ignore the bitrate when it's not set.

Also a small example of how to use it.

### DIFF
--- a/picamera2/encoders/h264_encoder.py
+++ b/picamera2/encoders/h264_encoder.py
@@ -4,6 +4,8 @@ from math import sqrt
 
 from v4l2 import (V4L2_CID_MPEG_VIDEO_H264_I_PERIOD,
                   V4L2_CID_MPEG_VIDEO_H264_LEVEL,
+                  V4L2_CID_MPEG_VIDEO_H264_MAX_QP,
+                  V4L2_CID_MPEG_VIDEO_H264_MIN_QP,
                   V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER,
                   V4L2_MPEG_VIDEO_H264_LEVEL_4_1,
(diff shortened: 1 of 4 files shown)


picamera2/encoders/h264_encoder.py
```diff
@@ -26,18 +28,24 @@ def __init__(self, bitrate=None, repeat=True, iperiod=None, framerate=None, enab
...
+        self.repeat = repeat
+        self.qp = qp
         # The framerate can be reported in the sequence headers if enable_sps_framerate is set,
         # but there's no guarantee that frames will be delivered to the codec at that rate!
         self._framerate = framerate
         self._enable_framerate = enable_sps_framerate
 
     def _start(self):
+        if self.iperiod is not None:
+            self._controls += [(V4L2_CID_MPEG_VIDEO_H264_I_PERIOD, self.iperiod)]
```
chrisruk, line 45: Just thinking because you've moved this code from __init__ to _start, won't self._controls, be concatenated to each time you start the encoder?

davidplowman: Indeed, and I've got it wrong for all the controls in there! I think I was wanting folks to be able to set or change these things after initialising the encoder object, so long as they do it before starting (or re-starting after stopping) it. Maybe putting `self._controls = []` at the top would do, or might we need anything more sophisticated?

chrisruk:
Looks good, can merge, just noticed a small typo in the first commit title, a bracket, not sure if we need to alter that or not.

davidplowman:
Groan. No, I'll fix that. Thanks for spotting it!

PR #599, opened 2023-03-02
More flexibility in managing and running paritcular lists of tests

davidplowman:
Two commits here: the first changes the run_tests.py script to accept a comma-separated list of files (each containing a collection of tests, as we currently have). The existing default behaviour is maintained.

The second commits adds a first test for "libcamera behaviour", i.e. stuff that we should run to test our libcamera implementation, but there's no point running on every Picamera2 commit.

### DIFF
--- a/tests/switch_mode_exposure.py
+++ b/tests/switch_mode_exposure.py
@@ -1,0 +1,42 @@
+import time
+
+from picamera2 import Picamera2
+
+picam2 = Picamera2()
+if 'imx708' in picam2.camera_properties['Model']:
+    print("SKIPPED (imx708)")
+    quit()
+
+# Select the smallest full FoV mode for the preview.
+preview_mode = None
+for mode in picam2.sensor_modes:
+    if mode['crop_limits'][:2] == (0, 0) and \
+       (preview_mode is None or mode['size'][0] < preview_mode['size'][0]):
+        preview_mode = mode
+print("Preview mode:", preview_mode)
+
+preview_config = picam2.create_preview_configuration(raw=preview_mode)
+still_config = picam2.create_still_configuration(buffer_count=2)
+
+picam2.start()
+time.sleep(1)
+preview_metadata = picam2.capture_metadata()
+
+picam2.switch_mode(still_config)
+time.sleep(1)
+still_metadata = picam2.capture_metadata()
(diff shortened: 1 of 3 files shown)


chrisruk:
Looks good, I've just been running the example, I seem to get the following though each time -
```
Exposures 59994 59994
Gain 2.540942907333374 3.002932548522949
ERROR: unreasonable change in gain, 2.540942907333374 vs 3.002932548522949
```

It seems quite overcast here, but wondered if the camera should be covered potentially.

davidplowman:
Might must be worth checking you're up to date with raspberrypi/libcamera next branch. There was indeed a bug in there that I fixed and added this test for it!

chrisruk:
Ah cheers, I'll recompile / retest, thanks

chrisruk:
Strangely I still seem to get this issue using the pi camera 3, wide and normal

It says -
[0:00:34.728581923] [1468]  INFO Camera camera_manager.cpp:299 libcamera v0.0.0+4212-923f5d70

Which seems to match up with the version I just pulled and re-compiled, with latest git commit has - 923f5d707bb760bd3e724b3373568fa88c68454f

chrisruk:
Just found it works with this camera though -
[0:00:37.899923439] [1458]  INFO RPI raspberrypi.cpp:1476 Registered camera /base/soc/i2c0mux/i2c@1/ov5647@36 to Unicam device /dev/media0 and ISP device /dev/media3

davidplowman:
Well, it might point to something that genuinely requires investigation. It works on my HQ cam, but I've not tested any others yet.

davidplowman:
Ah I've remembered. There this thing with the v3 camera where exposures in the binned mode are exactly half what they are in the full readout mode. So we'd have to take that into account, though then we start to have problems with not being in the same part of the exposure profile (so the exposure and gain get shared out differently, and possibly even clamped).

So I'm thinking this test just won't work for a v3 camera. Maybe I should exclude it?

davidplowman:
I've stopped the new tests from running on the v3 camera - as the exposure is actually expected to change on account of the different mode sensitivities.

chrisruk:
Cool, yeah skipping sounds a good idea, will merge now

PR #603, opened 2023-03-06
Add the ability to stop the camera quickly

davidplowman:
Currently, stopping the camera requires you to wait for a frame to arrive which can be slow when taking long exposures. This set of commits implements this feature by allowing the event loop to be "prodded" immediately so that prcoess_requests() runs straight away, allowing us to stop the camera.

Although there are quite a few commits, each is trying to make just a small self-contained change towards the end goal. The commits are:

1. Currently process_requests() will bail out if no new requests have arrived. This first change merely makes it run through successfully (doing nothing) if that happens.
2. When a list of "jobs" has been queued up for process_requests() to execute, they're ignored unless there are completed requests available. This change removes this restriction (as stopping the camera will be something we want to happen even if there are no requests). As a side effect, all the functions that jobs currently call need to check for the presence of requests and say that they must "try again later" in that case.
3. This commit causes the event loop (therefore process_requests()) to run immediately if it looks like the job might complete immediately (i.e. there's a completed request waiting already, and the job is the only one in the queue). We also add a flag to force the "prod the event loop immediately" behaviour as we'll need that in due course.
4. The mechanism added in 3 makes the existing mechanism for doing this (the execute_or_dispatch function) redundant, so I've removed that in favour of the new version.
5. Finally, all the jobs that start by stopping the camera (including anything that starts with a mode switch) are updated to prod the event loop immediately, so they should happen instantly.
6. This commit is a test that checks we can indeed stop the camera instantly even when a long exposure is running.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1060,7 +1060,7 @@ class Picamera2:
         # to stop, otherwise we can stop directly. When running a proper Qt app, _preview
         # is unset because we expect this code to be running the the Qt thread.
         if self._preview is not None and self._event_loop_running:
-            self.dispatch_functions([self.stop_], wait=True)
+            self.dispatch_functions([self.stop_], wait=True, immediate=True)
         else:
             self.stop_()
 
@@ -1216,7 +1216,7 @@ class Picamera2:
     def switch_mode(self, camera_config, wait=None, signal_function=None):
         """Switch the camera into another mode given by the camera_config."""
         functions = [partial(self.switch_mode_, camera_config)]
-        return self.dispatch_functions(functions, wait, signal_function)
+        return self.dispatch_functions(functions, wait, signal_function, immediate=True)
 
     def switch_mode_and_capture_file(self, camera_config, file_output, name="main", format=None,
                                      wait=None, signal_function=None):
@@ -1233,7 +1233,7 @@ class Picamera2:
 
         functions = [partial(self.switch_mode_, camera_config),
                      partial(capture_and_switch_back_, self, file_output, preview_config, format)]
-        return self.dispatch_functions(functions, wait, signal_function)
+        return self.dispatch_functions(functions, wait, signal_function, immediate=True)
 
     def switch_mode_and_capture_request(self, camera_config, wait=None, signal_function=None):
         """Switch the camera into a new (capture) mode and capture a request, then switch back.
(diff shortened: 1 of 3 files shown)


davidplowman:
I wonder if the test failure is caused by the v2 camera not being able to do a slow enough framerate? I will investigate.

davidplowman:
The max frame time on a v2 sensor for the mode that gets selected is 5.88 seconds, so I've reduced upped the 0.1 fps to 0.2. Actually even that required a kernel/driver update, the limit was only about 0.6s per frame with the versions that were installed!

chrisruk:
Just been looking through this @davidplowman , looks good to me.

Very minor question with `def capture_metadata_(self)` am I right in thinking that method isn't designed to be called by the user.  If so, I think it should be `def _capture_metadata(self)`.

davidplowman:
This is an interesting point. There are quite a few functions that I wouldn't expect most people to call, but they are available for people who know what they are doing. The relevant bit of the manual is section 8.2.2 - functions that can safely be called within the event loop normally have the underscore on the end. So it's not that they shouldn't be available, but that they are "special" in some way.

Not sure if there are better conventions for this sort of thing...?

chrisruk:
I've just been looking at - https://peps.python.org/pep-0008/ 

I found the following section -

```
_single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.

single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g
```

In the python REPL say you do:
```
class Test:
    def a(self):
        print("hi")
    def b(self):
        print("hi")
    def _c(self):
        print("hi")

t = Test()
```

Then you type `t.<TAB>` it will only list functions a and b, but hide _c, as far as I understand because it's for internal use.
However it doesn't actually prevent you from calling `t._c()`.    (You can get it to list underscored functions/variables though, by doing `t._<TAB>`).

davidplowman:
So it sounds like maybe the leading `_` is better? It might still be nice to have some other way to label this set of functions, though. Maybe a trailing `_` as well! Or perhaps not...

chrisruk:
Yeah I think the leading `_` is probably better. I did wonder if you have a trailing `_` too, it looks a bit like the convention used by Python's 'magic methods', such as `__str__` .

Should I merge this pull request now though, and if we want to change the naming styles maybe that could be done in a different pull req.

davidplowman:
I agree, an underscore on the end as well looks weird!

But you're right, anything like that wants to be a separate commit, there's certainly no need to muddle it up with anything here! So I'm good to merge...

PR #611, opened 2023-03-09
Create capture_images_by_key.py

Petros626:
I just created a simple example to take pictures with the keyboard. If there's any issue on performance or anything else, tell me.

Hope it helps some people, if you searched something like this.

### DIFF
--- a/examples/capture_images_by_key.py
+++ b/examples/capture_images_by_key.py
@@ -4,12 +4,13 @@
 # The image release is activated with the button 'p' and with 'q' the script is stopped.
 
 from picamera2 import Picamera2, Preview
+
 from sys import stdin
 from keyboard import is_pressed
 from termios import tcflush, TCIOFLUSH
 from time import strftime
 
-key_flag = False  
+key_flag = False
 cam = Picamera2()
 cam.start_preview(Preview.QTGL)
 cam.start()
@@ -17,9 +18,9 @@ cam.start()
 try:
     while True:
         if is_pressed('p'):
-            if key_flag == False:
+            if key_flag is False:
                 key_flag = True
-            filename =  strftime("%Y%m%d-%H%M%S") + '.png'
+            filename = strftime("%Y%m%d-%H%M%S") + '.png'
             cam.capture_file(filename, format="png", wait=None)
             print("\rCaptured {} succesfully".format(filename))
         else:


Petros626:
Thanks for watching over, if you got any suggestions how to improve it let me know. How would a non-blocking look like

tvoverbeek:
Tricky depending on if you are running headless or from a terminal window on the desktop.
Search for packages like `keyboard` and `pynput` on PyPi.

tvoverbeek:
@Petros626 I do not have commit permission and am  a 'normal' user just like you. Check with @davidplowman and/or @chrisruk

davidplowman:
Hi, sorry, it sounded from the above like you were still wanting to make the keypress non-blocking? I would agree that this would seem like the behaviour people would mostly expect? But please let me know.

Thanks for adding the sign-off. Could you squash everything into a single commit - wherever we can (it's not always possible of course), we prefer to have single easy-to-understand commits rather than seeing all the trial and error that led to the finished version! Thanks.

Petros626:
@davidplowman Now it's fine and I squashed all the commits in one.

davidplowman:
Hi, I'll have a look at the test failure, it looks like an intermittent problem related to a previous commit.

Would you be OK to look at the flake8 complaints? I think they're probably very trivial. Thanks!

Petros626:
Sure I guess/hope the linting errors, should be solved. You squash the two commits or should I do it?

davidplowman:
I'm happy to do it, but can you supply an email address I can use in a "Signed-off-by:" line? Thanks! (Or alternatively, you could squash them and add your sign-off.)

Petros626:
> I'm happy to do it, but can you supply an email address I can use in a "Signed-off-by:" line? Thanks! (Or alternatively, you could squash them and add your sign-off.)

Still facing some errors, but I thought I solved them, maybe you can show how to do this correctly. The e-mail address is: petros.katsoulakos@web.de

davidplowman:
Yes, isort frequently baffles me with its complaints. Often it's easiest just to run "python3 -m isort <problem-file>" and it fixes it for you! For the format string, it seems not to like the `{}`. Maybe use named-parameter sytle, or perhaps an f-string.

davidplowman:
Actually let me do all that for you...

PR #637, opened 2023-03-24
Fix problem passing modes from the sensor_modes list as the raw stream

davidplowman:
The intention is that an application can pick a sensor mode from the list returned by picam2.sensor_modes and pass that directly as the raw stream to the "create configuration" functions. For some reason this was enabled only for preview configurations so here we add it to the others.


davidplowman:
A relatively minor thing, but still not really sure how we managed to overlook it!

PR #649, opened 2023-04-04
Reduce the about of CMA memory required for typical use cases

davidplowman:
This commit points us to use one of the new pipeline configuration files (rpi_apps.yaml), which should save memory particularly in scenarios where we have a small number of high resolution buffers (so typically, still image capture).

One consequence of this is that if you don't request a raw stream, then you don't really get enough buffers to run sensibly, so all the create_xxx_configuration functions are updated to request a raw stream by default, even when the user didn't explicitly want one. There should be no downside to this.

### DIFF
--- a/picamera2/__init__.py
+++ b/picamera2/__init__.py
@@ -1,3 +1,5 @@
+import os
+
 import libcamera
 
 from .configuration import CameraConfiguration, StreamConfiguration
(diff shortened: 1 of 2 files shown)


davidplowman:
Hmm, all the dual camera stuff seems to have failed. Better have a look at that!

tvoverbeek:
PR title: about = amount ?

davidplowman:
> PR title: about = amount ?

Indeed, thank you!!

davidplowman:
@chrisruk Hi Chris, could you maybe give this a try as well just to check that nothing untoward happens for you either? Also make sure that you've got the latest version of raspberrypi/libcamera. Thanks!!

chrisruk:
@davidplowman Sure, will test this morning

chrisruk:
Just did `apt update` and `apt upgrade`, so I now have the 64 bit kernel/32 bit userland.

And upgraded libcamera from the 'next' branch.

All but one test passed for me.

```
Running /home/pi/Repositories/picamera2/tests/stop_slow_framerate.py...         ERROR
         error: frame arrived too quickly 0.9642159938812256 seconds
```
I assume that's probably unrelated to this pull request though?

chrisruk:
The test that failed for me, fails when using the latest picamera2 'next' branch also.  So unrelated to this pull request.  That was the only test failed after running all tests twice.

davidplowman:
Probably, what sensor are you using?

chrisruk:
Looks like it's an ov5647 camera, will switch to one of the newer autofocus cameras and try again.

davidplowman:
That's still a bit strange, the ov5647 should be able to do up to 6 second exposures, and the test asks for 5 (well, 5 seconds per frame). Might be worth checking what `picam2.sensor_modes` reports, I guess it's possible we're falling into a mode that can't do this. But as you say, maybe we don't care about the ov5647 so much, good idea to try a different one.

chrisruk:
Just tried with the autofocus camera and ran all examples twice, which all passed :)

Will merge this now, if that's ok?

davidplowman:
Actually can we hold off, there are still one or two things niggling me!

PR #665, opened 2023-04-14
Example to capture a specific number of frames

chrisruk:
Captures a video file with a specific number of frames, rather than using a duration.

Tested the number of frames in the generated .h264 file, appeared correct using -
```
 ffprobe -v error -select_streams v:0 -count_packets -show_entries stream=nb_read_packets -of csv=p=0 test.h264
```


davidplowman:
Hi Chris, interesting example!

I suppose the only minor observation is that you _could_ (if the requests are being handled in a "lumpy" way, maybe because the system is busy), get 2 requests coming through at once. The first one might cause the camera to stop, but then the second one would still get encoded.

Not sure how you would handle this... maybe count frames in the encoder output class?

chrisruk:
Thanks David, very good point, I've tweaked it to subclass the encoder in the example to avoid writing more frames than we want in there.

examples/capture_video_nframes.py
```diff
@@ -0,0 +1,38 @@
...
+
+encoder = H264EncoderStop(10000000)
+picam2.start_recording(encoder, 'test.h264')
+
+# Wait for 200 frames to be captured
+with encoder.cond:
+    encoder.cond.wait()
+
+# Necessary to ensure all frames are written
+time.sleep(1)
```
davidplowman, line 38: Just wondering if `picam2.stop_recording()` would ensure everything gets flushed? Then we could perhaps dispense with the sleep, which feels slightly awkward...

chrisruk: Strangely it doesn't, I just tried that now and got 199 frames written, I agree the sleep is awkward

davidplowman: That's unfortunate.  What about doing the counting in the encoder's `outputframe` method? By the end of that function the output has actually been given the frame, so hopefully it will get flushed before it stops?

chrisruk: Thanks, that did the trick :) I've moved to subclassing the FileOutput class instead and stopping that.

davidplowman: Cool, I'll let the tests run through and then merge it.

PR #678, opened 2023-04-20
Update opencv_face_detect.py

steveiliop56:
The problem with the `opencv_face_detect.py` is in that some systems and versions of opencv the frames will not update without the `cv2.waitKey(1)` function, which probably adds a very small delay so opencv has time to process the new frame. This change will not change any current functionality of the code but it will make it more reliable and offer an easy way to exit the program rather than spamming `Control + C` in the terminal.

**Signed-off-by:** Stavros Iliopoulos steveiliop56@gmail.com

### DIFF
--- a/examples/opencv_face_detect.py
+++ b/examples/opencv_face_detect.py
@@ -23,3 +23,5 @@ while True:
         cv2.rectangle(im, (x, y), (x + w, y + h), (0, 255, 0))
 
     cv2.imshow("Camera", im)
+    cv2.waitKey(1)
+    


davidplowman:
Hi, thanks for the PR. I'm happy to merge it, just a couple of teeny things:

* I wasn't quite sure whether you'd signed the commit message, or just the PR request? If you could sign the commit message that would be great. A short extra sentence or two saying what you fixed is always appreciated.
* flake8 seems to be complaining about a blank line with whitespace in it - could you maybe just remote that?

Thanks very much!

steveiliop56:
Ok I updated the script and changed the file to remove the problematic line. I am sorry but I wasn't able to sign the commit because I am using the GitHub editor. Is there anything else that I have to do?

davidplowman:
Hi, I've patched things up and merged this directly. Please let me know if it's not what you expected! Thanks.

steveiliop56:
Yeah it's totally fine thank you very much for you time.

Issue #684, opened 2023-04-25
[BUG] Encoders parameter 'quality' not functioning


cm-saiki-kyohei:
**Describe the bug**
The 'quality' parameter in the Encoders, as described in the Picamera2 Manual, does not appear to be functioning.

Reference: [Picamera2 Manual](https://datasheets.raspberrypi.com/camera/picamera2-manual.pdf) (p.41-42)

**To Reproduce**
I recorded videos with the following three quality conditions, but there did not seem to be any differences in quality between them:

```python
import time
from picamera2.encoders import H264Encoder, Quality
from picamera2 import Picamera2

picam2 = Picamera2()
video_config = picam2.create_video_configuration()
picam2.configure(video_config)
encoder = H264Encoder()

picam2.start_recording(encoder, 'low.h264', quality=Quality.VERY_LOW)
print(encoder.bitrate)
time.sleep(5)
picam2.stop_recording()

picam2.start_recording(encoder, 'medium.h264', quality=Quality.MEDIUM)
print(encoder.bitrate)
time.sleep(5)
picam2.stop_recording()

picam2.start_recording(encoder, 'high.h264', quality=Quality.VERY_HIGH)
print(encoder.bitrate)
time.sleep(5)
picam2.stop_recording()
```

**Expected behaviour**
The recorded videos should have noticeable differences in quality based on the specified 'quality' parameter.

**Console Output, Screenshots**
N/A

**Hardware(Environment information):**
- Raspberry Pi: Raspberry Pi 3 model B
- Camera: Freenove FNK0056
- Raspberry Pi OS: Debian GNU/Linux 11 (bullseye)
- Python: 3.9.2
- picamera2: 0.3.9

**Additional context**
The issue seems to be caused by the fact that the 'quality' argument is passed to the following function, but no processing is done within the function: [Relevant part in encoder.py](https://github.com/raspberrypi/picamera2/blob/d07a49197622d636bd993a5bf1de321c259abf47/picamera2/encoders/encoder.py#L240-L241)

davidplowman:
Hi, thanks for reporting that. Yes, this does seem to be broken. I'll fix it for the next release. In the meantime I think you would probably need to pass the bitrate explicitly as a workaround.

PR #685, opened 2023-04-25
H264 encoder quality was having no effect on bitrate

davidplowman:
The quality parameter was being ignored. Code also tweaked so that setting an explicit quality causes any previous bitrate to be overwritten, whereas omitting the quality will leave it alone (or default to medium quality if there was no bitrate set).

A test is also added to check that the bitrate does indeed change as described.


davidplowman:
Fixes https://github.com/raspberrypi/picamera2/issues/684

PR #709, opened 2023-05-31
switch_mode_and_capture_ improvements

davidplowman:
We should probably be more careful to check that we really do have a proper request after the switch mode, given that we now let process_requests run even when there aren't any.


davidplowman:
I believe the test failure here is fixed by https://github.com/raspberrypi/picamera2/pull/707.

PR #764, opened 2023-07-29
Enable atomic commits for DRM rendering

peytonicmaster6:
Signed-off: Peyton Howe <peyton.howe@bellsouth.net>

### DIFF
--- a/picamera2/previews/drm_preview.py
+++ b/picamera2/previews/drm_preview.py
@@ -193,13 +193,13 @@ class DrmPreview(NullPreview):
                 self.drmfbs[fb] = drmfb
 
             drmfb = self.drmfbs[fb]
-            self.crtc.set_plane(self.plane, drmfb, x, y, w, h, 0, 0, width, height)
-            # An "atomic commit" would probably be better, but I can't get this to work...
-            # ctx = pykms.AtomicReq(self.card)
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks for this PR. Is is really just using `commit_sync` as opposed to `commit` that makes it work? I never would have guessed...!

One thing I wonder about is whether the code should be reorganised a bit so that the image and overlay are added to the same atomic commit? Presumably that's where the benefit would be, in that we'd only be waiting for one vsync for both as opposed to two. What do you think?

peytonicmaster6:
I was surprised that `commit_sync` worked while `commit` didn't as they both call drmAtomicModeCommit(). The only difference is that commit passes in the `DRM_MODE_PAGE_FLIP_EVENT` and `DRM_MODE_ATOMIC_NONBLOCK` flags, while `commit_sync` doesn't pass in any flags.

Using the same atomic commit for the image and the overlay is definitely the way to go, otherwise the overlay will update at a faster refresh rate than the camera feed.

davidplowman:
We're proposing to replace this PR by https://github.com/raspberrypi/picamera2/pull/925, which should do the same thing.

davidplowman:
Closing this as we've merged the "revised version". Hope that's all good!

Issue #71, opened 2022-04-20
Worth adding custom EXIF to capture_file?


barryhunter:
I would like to write some additional EXIF tags to the captured .jpg file (specifically GEO - lat/long etc from a GPS, and possibly some other arbiary notes etc), rather than writing the jpeg, then having to add the data afterwards. 

Of course, could just grab a raw image (with  `make_image`) and then write the file, but as the library already has the `capture_file` method, wondering if would be interest in adding support for writing arbitrary EXIF data? Seems would be a common use case, and would avoid users having to duplicate the code in `CompletedRequest.save()`

Propose to extend the `save()` method, something like:
```
def save(self, name, filename, exif_dict={})
...
# add some extra tags to the user supplied data
exif_dict["0th"][piexif.ImageIFD.Make] = "Raspberry Pi"
...
exif_dict["Exif"][piexif.ExifIFD.ISOSpeedRatings] = int(total_gain * 100)
exif = piexif.dump(exif_dict)
```
and of course add to `capture_file()` (and `capture_file_`) so the data is passed though

Just unsure if would add before/after `signal_function`, suspect before, (because most users won't use `signal_function`) 
`def capture_file(self, filename, name="main", wait=True, exif_dict={}, signal_function=signal_event):
`

Then users would 
```
gps_ifd = {
        piexif.GPSIFD.GPSVersionID: (2, 0, 0, 0),
        #...
}
exif_dict = {"GPS": gps_ifd}
picam2.capture_file("test.jpg", exif_dict = exif)
```
(above code is just proposal, not tested yet!) 

I'd be willing to do this as a PR. Just checking if this the right direction first.

davidplowman:
Hi, thanks for suggesting this. Yes, I think it sounds like a good idea. The one thing I want to avoid is to have lots of exif-handling code of our own (like in _libcamera-apps_ where the whole exif thing is unbelievably painful), but in the Python world the situation is much better. I'd certainly be happy with something where you can optionally supply a dict of tags using `piexif`, and these just get passed though, more or less without touching anything.

So yes, please do submit a PR. Make sure to update the `switch_mode_and_capture_file` methods too. Also, you may have seen it, but I've tried to outline some guidance for contributions [here](https://github.com/raspberrypi/picamera2#contributing) (nothing unreasonable, I hope).

davidplowman:
Hi again, I'm going to close this one for the time being. If you'd like to submit a PR as discussed, please feel free to do so. Thanks!

PR #779, opened 2023-08-24
Add user defined exif_data when capturing files

Amudtogal:
When capturing to EXIF-capable files, the user can provide the
dictionary `exif_data` which is merged with the auto-generated EXIF-data
before writing into the image file.

Changes affect the `capture_file_` method and all derivatives.

The commit is based on this old discussion #71 and adds the requested
functionality.

Documentation in docstrings has been provided where applicable.

As I have no access to a working picamera environment right now, I can not test
the functionality.


picamera2/request.py
```diff
@@ -269,7 +277,10 @@ def save(self, img, metadata, file_output, format=None):
                 exif_ifd = {piexif.ExifIFD.DateTimeOriginal: datetime_now,
                             piexif.ExifIFD.ExposureTime: (metadata["ExposureTime"], 1000000),
                             piexif.ExifIFD.ISOSpeedRatings: int(total_gain * 100)}
-                exif = piexif.dump({"0th": zero_ifd, "Exif": exif_ifd})
+                exif_dict = {"0th": zero_ifd, "Exif": exif_ifd}
+                # merge user provided exif data, overwriting the defaults
+                exif_dict = exif_dict | exif_data
```
Amudtogal, line 282: This line works in Python 3.9+, for Python 3.5+ instead, we would do: `exif_dict = {**exif_dict, **exif_data}`, but I wasn't sure which Python version this is based on...

PR #809, opened 2023-10-02
DMA heap allocator

will-v-pi:
Add a DMA heap allocator, to allow the output buffers to be cached. This can give a large performance improvement when the MappedArray is then used by the application, and also elsewhere whenever the buffers are used.


davidplowman:
@will-v-pi Thanks. I've updated the libcamera installation so the tests should stand a slightly higher chance of working next time. Do you want to have a look at the flake8 complaints? I did try to tidy them all up recently (though it does look like there might be a few still lurking). Thanks!

davidplowman:
All looks great to me! One small thing, should we include a small test of using the libcamera allocator as well? I don't think there's any reason a user would ever do this, but maybe adding a small test case into picamera2/tests would ensure we don't accidentally break it in some way?

PR #812, opened 2023-10-03
Fix configuration issues with USB cams

davidplowman:
The new sensor configuration code doesn't work with USB cams. Make minimal changes just to stop things from falling over.


davidplowman:
@will-v-pi Do you want to push the "Rebase and merge" button if you're happy? Thanks!

PR #817, opened 2023-10-09
Make requests share a common lock

davidplowman:
This is necessary because all requests access common DMA allocator data structures.


davidplowman:
@will-v-pi Do you want to hit the rebase-and-merge button if you're happy that there isn't anything else... thanks!!

PR #819, opened 2023-10-10
Add a test that we can run with no raw buffers

davidplowman:
Otherwise this code path within our libcamera pipeline handler does not get tested.


davidplowman:
@naushir Do you want the dubious honour of pushing the rebase and merge button (if it "LGTY")?

PR #826, opened 2023-10-17
Sort the cameras alphabetically by id string

davidplowman:
This ensures the order will be deterministic on every reboot.


davidplowman:
@naushir @will-v-pi This is supposed to sort the cameras in the same order as libcamera-apps, the whole "cameras enumerated in a random order on every reboot" thing is a bit tiresome!

PR #828, opened 2023-10-19
FfmpegOutput was not stopping when there was an audio stream

davidplowman:
ffmpeg.wait() hangs indefinitely if there's an audio stream because no one is stopping it. Instead, give the video a few moments to flush itself out and then use ffmpeg.terminate().


davidplowman:
@naushir @will-v-pi Any objections?

PR #850, opened 2023-11-08
Make camera configuration objects work properly with sensor configuration

davidplowman:
You can now set the `sensor.output_size` and `sensor.bit_depth` in the camera config object. You can also set it to `None` so that any raw stream parameters will take precendence.

### DIFF
--- a/picamera2/configuration.py
+++ b/picamera2/configuration.py
@@ -88,10 +88,17 @@ class StreamConfiguration(Configuration):
     _FORWARD_FIELDS = {}
 
 
+class SensorConfiguration(Configuration):
+    _ALLOWED_FIELDS = ("output_size", "bit_depth")
+    _FIELD_CLASS_MAP = {}
+    _FORWARD_FIELDS = {}
+
+
 class CameraConfiguration(Configuration):
     _ALLOWED_FIELDS = ("use_case", "buffer_count", "transform", "display", "encode", "colour_space",
                        "controls", "main", "lores", "raw", "queue", "sensor")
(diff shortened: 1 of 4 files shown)


siranshen:
0.3.14 is broken due to this..

davidplowman:
> 0.3.14 is broken due to this, and apt install can't find 0.3.12 any more. It would be great if the version can be reverted while the fix is ongoing.

0.3.12 won't work with the new version of libcamera and Linux kernel. Are you able to use the workaround where you clear the configuration object before reusing it? (e.g. `picam2.still_configuration = picam2.create_still_configuration()` and then you do the rest of your configuration as before)

siranshen:
No... Is it possible to have both versions released and available for apt install?

davidplowman:
> No... Is it possible to have both versions released and available for apt install?

You would have to back out the kernel and libcamera, find versions that are compatible, and then install Picamera2 0.3.12 (which is available through pip, it just won't work any more). Perhaps you could to a pip install from this fix branch? Just remember to uninstall it once a fixed apt package becomes available.

siranshen:
Good point. I'll use pip in a virtual env. Thanks

davidplowman:
@naushir @will-v-pi Please shout if you see any acts of great stupidity in this PR...

Manuel-Angel-Es:
@davidplowman,

In the picamera2 module ver. 0.3.14-1 on my RPi4 I have replaced the original configuration.py and picamera2.py files with the new files.

I have subsequently modified my own software following your instructions.

Once the appropriate tests were carried out, everything worked correctly as expected.

Thank you very much for your work and the quick resolution of this problem.

Best regards

naushir:
I see nothing obviously wrong with this change.

PR #851, opened 2023-11-08
Timestamp capture

davidplowman:
This commit adds the ability to capture a request but where nothing was exposed before a given timestamp (which may be "now").

The second commit is an example/test for this functionality.

### DIFF
--- a/examples/timestamp_capture.py
+++ b/examples/timestamp_capture.py
@@ -1,0 +1,49 @@
+#!/usr/bin/python3
+
+import time
+
+from picamera2 import Picamera2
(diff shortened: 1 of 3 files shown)


davidplowman:
@naushir @will-v-pi Do we like this feature? I do get asked this kind of question from time to time, so I think it may be useful. I wonder a bit whether all the `capture_xxx` methods should allow it, but then again, most folks who care about this level of detail can probably use `capture_request`.

naushir:
I can see this being useful.  LGTM.

PR #855, opened 2023-11-09
Notify users when FFmpeg output processes exit

davidplowman:
An FFmpeg output process can end spontaneously, for example, if it's using a network connection that gets closed. When this happens, allow the user to get a notification so that they can decide what to do.


davidplowman:
@naushir @will-v-pi One more fix for a problem that folks have brought up from time to time...

naushir:
LGTM

PR #892, opened 2023-12-12
picamera2: Allow setting of stride

will-v-pi:
Allow setting of the stride in stream configuration, and default to 0 if not set to allow libcamera to set the correct stride.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -641,7 +641,7 @@ class Picamera2:
         """
         if updates is None:
             return None
-        valid = ("format", "size")
+        valid = ("format", "size", "stride")
         for key, value in updates.items():
             if isinstance(value, SensorFormat):
                 value = str(value)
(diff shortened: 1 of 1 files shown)


picamera2/picamera2.py
```diff
@@ -839,6 +839,10 @@ def _update_libcamera_stream_config(libcamera_stream_config, stream_config, buff
         libcamera_stream_config.size = libcamera.Size(stream_config["size"][0], stream_config["size"][1])
         libcamera_stream_config.pixel_format = libcamera.PixelFormat(stream_config["format"])
         libcamera_stream_config.buffer_count = buffer_count
+        if stream_config.get("stride") is not None:
+            libcamera_stream_config.stride = stream_config["stride"]
+        else:
+            libcamera_stream_config.stride = 0
```
davidplowman, line 845: Would `libcamera_stream_config.stride = stream_config.get("stride", 0)` be the same?

davidplowman:
Thanks for this Will - are we happy that is doesn't rely on any other libcamera/libpisp changes? (The tests all pass, so presumably not...)

picamera2/picamera2.py
will-v-pi: Oh yes, that'll be cleaner, I'll change that and push a new commit

will-v-pi:
It does rely on a libpisp and libcamera change, but that's only to get it to actually set the stride, so worth merging it now in anticipation of those changes being merged, as once the libcamera change is merged it'll break picamera2 (because it's accidentally setting an incorrect stride, which is currently ignored but if it's not ignored then it'll break). Without those changes it'll just do nothing when you set the stride, it won't break or anything.

davidplowman:
The tests seem to have got very unhappy, do we know what that's about? I can't believe that teeny tweak would have caused it...!

will-v-pi:
It seems the stream_config sometimes has the stride set to None, so that was causing things to break - I've gone back to the original code which guarded against that case, but add a comment to say why

PR #894, opened 2023-12-13
Add Persistent Allocator

will-v-pi:
Add a variant of the DmaAllocator where the buffers are not deallocated after switching mode, allowing you to have multiple sets of buffers stored which can be switched between based on the camera configuration. It selects which buffers to use for a given configuration based off the use_case parameter, so this must be set uniquely for each configuration that will be used with the allocator.

### DIFF
--- a/examples/switch_mode_persist.py
+++ b/examples/switch_mode_persist.py
@@ -1,0 +1,49 @@
+#!/usr/bin/python3
+
+# Switch between modes without reallocating buffers.
+
+import time
+
+from picamera2 import Picamera2
+from picamera2.allocators import PersistentAllocator
+
+picam2 = Picamera2(allocator=PersistentAllocator())
+
+# Create preview configuration
+preview_config = picam2.create_preview_configuration()
+picam2.configure(preview_config)
+
+# Create full resolution preview configuration
+full_res_config = picam2.create_preview_configuration(
+    main={"size": picam2.sensor_resolution}, buffer_count=2, use_case="full_res_preview"
+)
+picam2.configure(full_res_config)
+
+# Create still configuration
+still_config = picam2.create_still_configuration()
+picam2.configure(still_config)
+
+# Use preview configuration
+picam2.configure(preview_config)
+picam2.start()
+time.sleep(2)
+picam2.stop()
+
+# Use other configuration
+picam2.configure(full_res_config)
+picam2.start()
+time.sleep(2)
+picam2.stop()
+
+# Back to preview
+picam2.configure(preview_config)
+picam2.start()
+# Delete other buffers to free up space
+picam2.allocator.deallocate("full_res_preview")
+time.sleep(2)
+
+# Capture image with still configuration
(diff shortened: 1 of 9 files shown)


examples/switch_mode_persist.py
```diff
@@ -0,0 +1,57 @@
...
+picam2.start_preview(Preview.QTGL)
+
+# Create preview configuration
+preview_config = picam2.create_preview_configuration()
+picam2.configure(preview_config)
+
+size = picam2.sensor_resolution
+# GPU won't digest images wider than 4096 on a Pi 4.
+if size[0] > 4096:
+    height = size[1] * 4096 // size[0]
```
davidplowman, line 21: I wonder if we should leave this out if the test isn't displaying anything - only because it slightly obscures the main point of the example. What do you think?

examples/switch_mode_persist.py
```diff
@@ -0,0 +1,57 @@
...
+    height -= height % 2
+    size = (4096, height)
+
+# Create other configuration
+other_config = picam2.create_preview_configuration(main={"size": size}, buffer_count=2)
+other_config["use_case"] = "other"
```
davidplowman, line 27: Ah, OK, so the buffers are accessed through the "use_case" name. I wonder about allowing `use_case="other"` in the `create_xxx_configuration` functions. (Though that does clutter them up even more, so I'm not sure...).

I always thought the "use_case" might be a useful concept, but in the event I never found a use for it, so I suppose that makes it quite a good solution here.

examples/switch_mode_persist.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Switch between modes without reallocating buffers.
+
+import time
+
+from picamera2 import Picamera2, Preview
+from picamera2.allocators import PersistentAllocator
+
+picam2 = Picamera2()
+picam2.allocator = PersistentAllocator()
```
davidplowman, line 11: Super minor, would an `allocator` parameter to the constructor be tidy?

`picam2 = Picamera2(allocator=PersistentAllocator())`

examples/switch_mode_persist.py
will-v-pi: The test does display this, so it probably needs to stay in? It's in the other switch mode examples too

examples/switch_mode_persist.py
will-v-pi: I can add `use_case="xxx"` arguments to the `create_xxx_configuration` functions - it does add a little bit more clutter, but I think that does sound like a good idea, to make it easier to use

examples/switch_mode_persist.py
will-v-pi: Yep, I'll add that

davidplowman:
Looks very cool to me! Before I merge, are we happy the we're taking care of the risk of leaks in any of those "persistent" allocations, maybe in configurations that aren't currently being used?

PR #895, opened 2023-12-13
Add stereo preview example

will-v-pi:
Add an example demonstrating the use of stride, where the stride is set to twice the image width to allow an image from a second camera to be copied into the empty space, creating a stereo image

### DIFF
--- a/examples/stereo_preview.py
+++ b/examples/stereo_preview.py
@@ -1,0 +1,69 @@
+#!/usr/bin/python3
+
+import time
+
+from picamera2 import MappedArray, Picamera2, libcamera
+
+cam2_request = None
+
+
+def pre_callback(request):
+    # Set the size, to make preview window and MappedArray remapping work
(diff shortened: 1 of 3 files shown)


6by9:
On vc4 at least we could allow the selection API to set the compose rectangle to cover the part of the double sized buffer that the ISP is meant to write into. The same buffer could therefore be written by 2 instances of libcamera (one for each eye). The offset will need to be aligned based on the ISP constraints.

I have no idea with regard how to pass that information through libcamera in order to request this though!

davidplowman:
@will-v-pi There was some kernel update required which is why we didn't merge this immediately, do we think that might all be sorted now? Perhaps it's worth rebasing and re-pushing...?

davidplowman:
This has now been rebased and merged here: https://github.com/raspberrypi/picamera2/pull/1146

Sorry it took so long!!

Issue #913, opened 2024-01-06
set_overlay() for all outputs


MazrimT:
**Please only include one suggestion or features request per issue!**

**Is your feature request related to a problem? Please describe.**
The set_overlay() function only seems to work for previews and not for any actual output.
This is not clearly documented and also feels illogical to new user that it would not just generally be an overlay no matter the output.

**Describe the solution you'd like**
overlays should be taken into account in all outputs, wether it's just a file or FileOutput or FfmpegOutput is used.

**Describe alternatives you've considered**
in ffmpeg in general you can add things like overlays, so being able to specify more inputs to a single ffmpegoutput in ffmpeg format would work, but currently ffmpegoutput does not seem to work with input in the command.

**Additional context**

davidplowman:
Yes, the overlay feature is very deliberately only something for use on preview windows, for example for doing little bits of UI such as crosshairs, buttons or rectangles. It's not intended to be part of the final image.

If you want something on the final image, I'd recommend using the `pre_callback` to draw text or other graphics onto your images. There are examples of this kind of thing, for example [here](https://github.com/raspberrypi/picamera2/blob/main/examples/timestamped_video.py).

MazrimT:
Thank you for pointing me in the right direction!

PR #920, opened 2024-01-11
Adds example for overlay_png

MazrimT:
Just an example file with how you can add a "overlay" of a png file to a video as discussed in issue https://github.com/raspberrypi/picamera2/issues/913


davidplowman:
Hi, thanks for this example, I think it looks good - and thank you for taking the time to create a new example for us!

One slight misgiving I have is that it's loading the new overlay (i.e. doing filesystem and decompression activity) in the pre_callback, which runs in the camera thread. Mostly I try to discourage this as it increases the risk of dropping camera frames.

I wonder whether if would be worth using a separate thread which just sits there in the background and updates the overlay_png at the correct moment, and the pre_callback, running asynchronously to this, simply "samples" the most recently loaded one. Any resizing of the overlay could in principle be done outside the camera thread, too. What do you think?

MazrimT:
@davidplowman That is a great idea, I updated to do just that.

davidplowman:
Thanks for the update - could you maybe squash the 3 commits into a single one? I always think that in a public repo it doesn't usually help other users to "see the development process"! If you haven't done so, could you ensure the commit is signed-off according to the [guidelines](https://github.com/raspberrypi/picamera2#contributing).

I had one final thing that I thought might be worth investigating

I was wondering a little bit whether we should try to combine the overlay using `cv2.addWeighted`. It might do all the alpha blending for us (so no need to handle the alpha channel separately). It might also give us better performance - quite a few OpenCV functions are good about using multiple threads on images to speed things up.

But let me know what you want to do on that. Thanks again!

MazrimT:
Have a much better idea for this, will open new when ready.

PR #925, opened 2024-01-15
Use atomic commits for DRM rendering

will-v-pi:
Use atomic commits for DRM rendering, and also add support for using pure python kms module if available, instead of current C++ bindings

### DIFF
--- a/picamera2/previews/drm_preview.py
+++ b/picamera2/previews/drm_preview.py
@@ -1,8 +1,15 @@
+import gc
 import mmap
 import threading
 
 import numpy as np
-import pykms
+
+try:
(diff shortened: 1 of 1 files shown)


davidplowman:
@will-v-pi Thanks for this! LGTM. Are you happy to go ahead and merge?

It seems a bit annoying that we have to invoke the garbage collector explicitly, this does seem to be a bit of a sore point with Python sometimes.

will-v-pi:
Yeah, I'm happy to merge - with the garbage collection, I added that as the card only seems to get garbage collected once a new preview is started rather than when a preview is stopped, which was causing issues in the drm stop restart test

PR #928, opened 2024-01-21
Add real-time object detection example for Raspberry Pi 4

ranizouaoui:
### What does this PR do?
This pull request adds a real-time object detection example for Raspberry Pi 4 using TensorFlow Lite with metadata format. The example leverages OpenCV, MediaPipe, and Raspberry Pi OS (64-bit).

### How to Test
1. Ensure you are using a Raspberry Pi 4 with Raspberry Pi OS (64-bit).
2. Install the required dependencies by running the following commands:
   ```bash
   sudo rm /usr/lib/python3.11/EXTERNALLY-MANAGED
   sudo pip3 install opencv-python==4.8.1.78
   sudo pip3 install mediapipe==0.10.8
   ```
1. Convert your TensorFlow Lite model to metadata format following the steps mentioned in the README.
2. Navigate to the real-time-object-detection-rpi4 directory.
3. Run the real-time detection script:
   ```bash
   python3 real_time_detection_with_labels_on_raspberry4.py
   ```

### DIFF
--- a/examples/tensorflow/real-time-object-detection-rpi4/README.md
+++ b/examples/tensorflow/real-time-object-detection-rpi4/README.md
@@ -1,0 +1,39 @@
+# Real-time Object Detection on Raspberry Pi 4
+
+## Requirements
+
+- Raspberry Pi 4 with Raspberry Pi OS (64-bit)
+- Python 3.11
+
+## Dependencies
+
+Before running the code, make sure to install the following dependencies:
+
+```bash
+sudo rm /usr/lib/python3.11/EXTERNALLY-MANAGED
+sudo pip3 install opencv-python==4.8.1.78
+sudo pip3 install mediapipe==0.10.8
+```
+## TensorFlow Lite Model with Metadata Format
+
+This example uses a TensorFlow Lite model with metadata format. To convert your model to metadata format, follow these steps:
+
+Prepare a text file containing the labels for your model.
+Obtain the TFLite model.
+Use the following Colab file to convert your model to metadata format:
+<a href="https://colab.research.google.com/github/tensorflow/tensorflow/blob/master/tensorflow/lite/g3doc/models/convert/metadata_writer_tutorial.ipynb">Convert to Metadata Format - Colab File </a>
+
+# Running the Code
+
+To launch the code, execute the following commands:
+```bash
+# Navigate to the project directory
+cd path/to/picamera2/real-time-object-detection-rpi4
+
+# Run the real-time detection script
+python3 real_time_detection_with_labels_on_raspberry4.py
+
+```
+<blockquote> 
+<p dir="auto">Make sure to replace path/to/picamera2/ with the actual path to your picamera2 repository.</p>
+<blockquote>
--- a/examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
+++ b/examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
@@ -1,0 +1,154 @@
+
+import argparse
+import sys
+import time
+
+import cv2
+import mediapipe as mp
+
+from mediapipe.tasks import python
+from mediapipe.tasks.python import vision
+
+from utils import visualize
+from picamera2 import Picamera2
+
+# Global variables to calculate FPS
+COUNTER, FPS = 0, 0
+START_TIME = time.time()
+picam2 = Picamera2()
+picam2.preview_configuration.main.size = (640,480)
+picam2.preview_configuration.main.format = "RGB888"
+picam2.preview_configuration.align()
+picam2.configure("preview")
+picam2.start()
+
+def run(model: str, max_results: int, score_threshold: float, 
+        camera_id: int, width: int, height: int) -> None:
+  """Continuously run inference on images acquired from the camera.
+
+  Args:
+    model: Name of the TFLite object detection model.
+    max_results: Max number of detection results.
+    score_threshold: The score threshold of detection results.
+    camera_id: The camera id to be passed to OpenCV.
+    width: The width of the frame captured from the camera.
+    height: The height of the frame captured from the camera.
+  """
+
+
+
+  # Visualization parameters
+  row_size = 50  # pixels
+  left_margin = 24  # pixels
+  text_color = (0, 0, 0)  # black
+  font_size = 1
+  font_thickness = 1
+  fps_avg_frame_count = 10
+
+  detection_frame = None
+  detection_result_list = []
+
+  def save_result(result: vision.ObjectDetectorResult, unused_output_image: mp.Image, timestamp_ms: int):
+      global FPS, COUNTER, START_TIME
+
+      # Calculate the FPS
+      if COUNTER % fps_avg_frame_count == 0:
+          FPS = fps_avg_frame_count / (time.time() - START_TIME)
+          START_TIME = time.time()
+
+      detection_result_list.append(result)
+      COUNTER += 1
+
+  # Initialize the object detection model
+  base_options = python.BaseOptions(model_asset_path=model)
+  options = vision.ObjectDetectorOptions(base_options=base_options,
+                                         running_mode=vision.RunningMode.LIVE_STREAM,
+                                         max_results=max_results, score_threshold=score_threshold,
+                                         result_callback=save_result)
+  detector = vision.ObjectDetector.create_from_options(options)
+
+
+  # Continuously capture images from the camera and run inference
+  while True:
+    im= picam2.capture_array()  
+#    success, image = cap.read()
+    image=cv2.resize(im,(640,480))
+    image = cv2.flip(image, -1)
+
+    # Convert the image from BGR to RGB as required by the TFLite model.
+    rgb_image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+    mp_image = mp.Image(image_format=mp.ImageFormat.SRGB, data=rgb_image)
+
+    # Run object detection using the model.
+    detector.detect_async(mp_image, time.time_ns() // 1_000_000)
+
+    # Show the FPS
+    fps_text = 'FPS = {:.1f}'.format(FPS)
+    text_location = (left_margin, row_size)
+    current_frame = image
+    cv2.putText(current_frame, fps_text, text_location, cv2.FONT_HERSHEY_DUPLEX,
+                font_size, text_color, font_thickness, cv2.LINE_AA)
+
+    if detection_result_list:
+        # print(detection_result_list)
+        current_frame = visualize(current_frame, detection_result_list[0])
+        detection_frame = current_frame
+        detection_result_list.clear()
+
+    if detection_frame is not None:
+        cv2.imshow('object_detection', detection_frame)
+
+    # Stop the program if the ESC key is pressed.
+    if cv2.waitKey(1) == 27:
+      break
+
+  detector.close()
+  cap.release()
+  cv2.destroyAllWindows()
+
+
+def main():
+  parser = argparse.ArgumentParser(
+      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument(
+      '--model',
+      help='Path of the object detection model.',
+      required=False,
+      default='efficientdet_lite0.tflite')
+  parser.add_argument(
+      '--maxResults',
+      help='Max number of detection results.',
+      required=False,
+      default=5)
+  parser.add_argument(
+      '--scoreThreshold',
+      help='The score threshold of detection results.',
+      required=False,
+      type=float,
+      default=0.25)
+  # Finding the camera ID can be very reliant on platform-dependent methods. 
+  # One common approach is to use the fact that camera IDs are usually indexed sequentially by the OS, starting from 0. 
+  # Here, we use OpenCV and create a VideoCapture object for each potential ID with 'cap = cv2.VideoCapture(i)'.
+  # If 'cap' is None or not 'cap.isOpened()', it indicates the camera ID is not available.
+  parser.add_argument(
+      '--cameraId', help='Id of camera.', required=False, type=int, default=0)
+  parser.add_argument(
+      '--frameWidth',
+      help='Width of frame to capture from camera.',
+      required=False,
+      type=int,
+      default=640)
+  parser.add_argument(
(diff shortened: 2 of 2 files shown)


ranizouaoui:
real-time object detection example for Raspberry Pi 4

davidplowman:
Hi, and thanks for this PR. In principle I'm happy to have more up to date TensorFlow example so would be happy to merge this. I had just a few questions about it, perhaps you could tell me what you think before we do that. I'll put some comments inline in your commits. Thanks!

examples/tensorflow/real-time-object-detection-rpi4/README.md
```diff
@@ -0,0 +1,39 @@
...
+- Python 3.11
+
+## Dependencies
+
+Before running the code, make sure to install the following dependencies:
+
+```bash
+sudo rm /usr/lib/python3.11/EXTERNALLY-MANAGED
+sudo pip3 install opencv-python==4.8.1.78
+sudo pip3 install mediapipe==0.10.8
```
davidplowman, line 15: I'm a bit hesitant about advising people to revert Bookworm's default behaviour which makes folks use venvs. Is it possible to do this in a venv?

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if detection_frame is not None:
+        cv2.imshow('object_detection', detection_frame)
+
+    # Stop the program if the ESC key is pressed.
+    if cv2.waitKey(1) == 27:
+      break
+
+  detector.close()
+  cap.release()
```
davidplowman, line 106: Is `cap` actually defined?

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+  parser.add_argument(
+      '--scoreThreshold',
+      help='The score threshold of detection results.',
+      required=False,
+      type=float,
+      default=0.25)
+  # Finding the camera ID can be very reliant on platform-dependent methods. 
+  # One common approach is to use the fact that camera IDs are usually indexed sequentially by the OS, starting from 0. 
+  # Here, we use OpenCV and create a VideoCapture object for each potential ID with 'cap = cv2.VideoCapture(i)'.
+  # If 'cap' is None or not 'cap.isOpened()', it indicates the camera ID is not available.
```
davidplowman, line 132: If we aren't using OpenCV's built-in video capture, perhaps we should remove these comments?

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+                                         max_results=max_results, score_threshold=score_threshold,
+                                         result_callback=save_result)
+  detector = vision.ObjectDetector.create_from_options(options)
+
+
+  # Continuously capture images from the camera and run inference
+  while True:
+    im= picam2.capture_array()  
+#    success, image = cap.read()
+    image=cv2.resize(im,(640,480))
```
davidplowman, line 75: Is the main image already 640x480? We might just be able to remove this line in that case!

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+    image = cv2.flip(image, -1)
```
davidplowman, line 76: Could we ask for a flipped image from the camera system in the first place? That's basically free so would be more efficient, though we obviously have to be sure it doesn't break the demo in some way.

examples/tensorflow/real-time-object-detection-rpi4/README.md
```diff
@@ -0,0 +1,39 @@
+# Real-time Object Detection on Raspberry Pi 4
+
+## Requirements
+
+- Raspberry Pi 4 with Raspberry Pi OS (64-bit)
+- Python 3.11
```
davidplowman, line 6: Presumably this would work on a Pi 5 as well? That would be something to test, though I appreciate you probably don't have one. Well, I can try it once everything else is sorted out!

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Convert the image from BGR to RGB as required by the TFLite model.
+    rgb_image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+    mp_image = mp.Image(image_format=mp.ImageFormat.SRGB, data=rgb_image)
+
+    # Run object detection using the model.
+    detector.detect_async(mp_image, time.time_ns() // 1_000_000)
```
davidplowman, line 83: I'm a bit curious as to what throttles this loop. Obviously the `picam2.capture_array()` limits it to the camera framerate, but presumably the object detection doesn't keep up with that - so what actually happens? Is the detector "queuing up" detection tasks that it doesn't keep up with (so the queue gets longer and longer), or does it at some point start waiting for detection tasks to finish (so becomes synchronous). Do you know how this will be handled?

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+  # Continuously capture images from the camera and run inference
+  while True:
+    im= picam2.capture_array()  
+#    success, image = cap.read()
+    image=cv2.resize(im,(640,480))
+    image = cv2.flip(image, -1)
+
+    # Convert the image from BGR to RGB as required by the TFLite model.
+    rgb_image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+    mp_image = mp.Image(image_format=mp.ImageFormat.SRGB, data=rgb_image)
```
davidplowman, line 80: I guess this is because Mediapipe/TensorFlow and OpenCV disagree about the order of R and B pixels? How very annoying! I'm trying to think if there's a way to avoid this. If we used the built-in Picamera2 preview then we could ask for and display the "low res" stream instead, but that's a bit of a change. Not sure!

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
```diff
@@ -0,0 +1,154 @@
...
+  options = vision.ObjectDetectorOptions(base_options=base_options,
+                                         running_mode=vision.RunningMode.LIVE_STREAM,
+                                         max_results=max_results, score_threshold=score_threshold,
+                                         result_callback=save_result)
+  detector = vision.ObjectDetector.create_from_options(options)
+
+
+  # Continuously capture images from the camera and run inference
+  while True:
+    im= picam2.capture_array()
```
davidplowman, line 73: Just as a general point, we run a flake8 pre-commit check which will enforce some (irritating!) tidy-ups on the code. Perhaps you could do this too so as to make the whole process quicker (read [here](https://github.com/raspberrypi/picamera2?tab=readme-ov-file#contributing)).

examples/tensorflow/real-time-object-detection-rpi4/README.md
ranizouaoui: Hello @davidplowman , I appreciate your concern about Bookworm's default behavior, and I understand the importance of encouraging the use of virtual environments (venvs). You can certainly perform the installation within a venv. Here's an updated set of instructions:

### Create and activate a virtual environment
```bash
python3.11 -m venv myenv
source myenv/bin/activate
```

### Install required packages within the virtual environment
```bash
pip install opencv-python==4.8.1.78
pip install mediapipe==0.10.8
```

<blockquote> 
<p dir="auto">Your subsequent commands and steps can be executed within the venv.</p>
</blockquote> 

<blockquote> 
<p dir="auto">Don't forget to deactivate the venv when you're done.</p>
</blockquote> 

```bash
deactivate
```

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: Yes, cap.release() exists in OpenCV 4.8.1.78. [It is a method that closes the video file or capturing device and frees up the resources that were being used by the device](https://stackoverflow.com/questions/48213499/whats-the-meaning-of-cv2-videocapture-release)[1](https://stackoverflow.com/questions/48213499/whats-the-meaning-of-cv2-videocapture-release). [It is the same as in previous versions of OpenCV, such as 4.5.1](https://stackoverflow.com/questions/67053118/cannot-release-videocapture-for-opencv)[2](https://stackoverflow.com/questions/67053118/cannot-release-videocapture-for-opencv). You can find more information about this method and other changes in OpenCV 4.8.1.78 in the [release notes](https://github.com/opencv/opencv-python/releases), the [official blog](https://opencv.org/blog/release/opencv-4-8-0/), or the [source code](https://sourceforge.net/projects/opencvlibrary/files/4.8.1/).

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: ### Thanks for your suggestion! You're right, if we're not using OpenCV's built-in video capture, these comments might be misleading.

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: ## Thanks for catching that! You're right, if the main image captured by 'picam2.capture_array()' is already 640x480, we can simplify the code by removing the resize operation:

   ```python
while True:
    im = picam2.capture_array()  
    # Assuming picam2.capture_array() already provides a 640x480 image
    image = im  # No need for cv2.resize in this case
   ```

examples/tensorflow/real-time-object-detection-rpi4/README.md
ranizouaoui: I noticed that the TensorFlow example code in the 'picamera 2' repository doesn't function properly on a Raspberry Pi 4 with Raspberry Pi OS (64-bit). In response, I've created my own version of the code and submitted this Pull Request to assist developers who may encounter this configuration in the future.

As for testing on Raspberry Pi 5, unfortunately, I currently don't have a Pi 5 card available. However, I'm committed to testing the code on a Raspberry Pi 5 as soon as I acquire one.

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: Thanks for your question! You've raised an important point about the potential mismatch between the camera frame rate and object detection speed. The current implementation utilizes asynchronous detection, and here's a brief explanation of how it works:

The `detector.detect_async()` function initiates the object detection process and continues with the loop without waiting for the detection to complete. It allows the code to keep up with the camera's frame rate.

The detector does not block the loop; instead, it may queue up detection tasks. If the queue becomes longer than the processing speed, it could lead to a backlog of tasks.

To address this, there might be a need to monitor the detection queue length and potentially introduce a mechanism to handle it, such as adjusting the camera frame rate or optimizing the detection process. 

This asynchronous approach aims to maintain responsiveness and leverage parallelism. If you have any suggestions or further questions on how to optimize this aspect, I'd be keen to explore them!

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: ### You've caught an interesting point! The conversion from BGR to RGB is indeed due to the difference in pixel order conventions between OpenCV and Mediapipe/TensorFlow.

I appreciate your consideration of alternatives. Using the built-in Picamera2 preview for the "low res" stream could be a potential solution, though it would involve a significant change. If you think it's worth exploring further, we can discuss the implications and benefits of such a modification. Feel free to share your thoughts on how we can handle this color order issue more efficiently!

examples/tensorflow/real-time-object-detection-rpi4/real_time_detection_with_labels_on_raspberry4.py
ranizouaoui: Thanks for the suggestion! I'll definitely incorporate the flake8 pre-commit check, as recommended in the contribution guide <a href="https://github.com/raspberrypi/picamera2?tab=readme-ov-file#contributing">here</a>. This will help ensure code consistency and streamline the review process. I appreciate your attention to code quality, and I'll make sure to follow the best practices outlined in the guide.

Signed-off-by: Rani Zouaoui <ranizouaouicontact@gmail.com>

examples/tensorflow/real-time-object-detection-rpi4/README.md
gwvr: I would urge caution around venvs.

I have had trouble running https://github.com/raspberrypi/picamera2/blob/main/examples/tensorflow/real_time.py because:
* With system python (outside of a venv) I can't install `tflite-runtime`
* In a venv PyQt/OpenCV is problematic - PyQt5 is not installable (`pip install PyQt5` on RPi 5 4GB results in OOM crash), so preview windows don't work and opencv imshow is badly behaved (usually fails to display)
* In a conda enviroment it is possible to run `real_time.py`, but only with a very heavy PyQt5 / OpenCV install, and the example doesn't seem complete (perhaps a README would help)

Ideally it would be possible to run all of the examples in a python virtualenv, but realistically I don't see how that's possible given the reliance on PyQt for preview windows.

PR #941, opened 2024-01-31
Drop first 20 frames before mse compare.

dksslq:
Avoid calculate mse of camera initialization and adaptation phase

### DIFF
--- a/examples/capture_motion.py
+++ b/examples/capture_motion.py
@@ -20,11 +20,13 @@ w, h = lsize
 prev = None
 encoding = False
 ltime = 0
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks for this PR. I agree that this is probably a sensible change to make so am happy to merge it. Could you just check that you've followed the [contribution guidelines](https://github.com/raspberrypi/picamera2?tab=readme-ov-file#contributing) - I think the only changes to make would be to sign off your commit, and to target the PR at the next branch. Thanks again!

PR #968, opened 2024-03-01
Deprecate use of the ScalerCropMaximum property

naushir:
This will be dropped by libcamera. Instead, use the maximum value of the ScalerCrop control, which stores an identical rectangle.

### DIFF
--- a/apps/app_full.py
+++ b/apps/app_full.py
@@ -7,7 +7,7 @@ from PyQt5.QtWidgets import (QApplication, QCheckBox, QComboBox,
                              QLineEdit, QPushButton, QSlider, QSpinBox,
                              QTabWidget, QVBoxLayout, QWidget)
 
-from picamera2 import Picamera2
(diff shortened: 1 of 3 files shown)


apps/app_full.py
```diff
@@ -92,7 +92,7 @@ def update_controls():
     global scaler_crop
 
     # Fix aspect ratio of the pan/zoom
-    full_img = picam2.camera_properties["ScalerCropMaximum"]
```
davidplowman, line 95: Actually (and sorry to be a pain...) could we do
```
_, full_img, _ = picam2.camera_controls['ScalerCrop']
```
I feel `camera_ctrl_info` should probably be made private at some point, and it would avoid people diving into the libcamera structures and having to do type conversions.

davidplowman:
Replaced by https://github.com/raspberrypi/picamera2/pull/973, so closing.

PR #973, opened 2024-03-06
Deprecate use of the ScalerCropMaximum property

davidplowman:
This will be dropped by libcamera. Instead, use the maximum value of the ScalerCrop control, which stores an identical rectangle.


davidplowman:
Replaces https://github.com/raspberrypi/picamera2/pull/968

PR #1039, opened 2024-05-25
add ffmpeg audio filter support

StephenGrable1:
enable -af flag to be passed through interface to support customizable audio configuration

### DIFF
--- a/picamera2/outputs/ffmpegoutput.py
+++ b/picamera2/outputs/ffmpegoutput.py
@@ -33,12 +33,13 @@ class FfmpegOutput(Output):
     """
 
     def __init__(self, output_filename, audio=False, audio_device="default", audio_sync=-0.3,
-                 audio_samplerate=48000, audio_codec="aac", audio_bitrate=128000, pts=None):
+                 audio_samplerate=48000, audio_codec="aac", audio_bitrate=128000, audio_filter='pan=stereo', pts=None):
         super().__init__(pts=pts)
         self.ffmpeg = None
         self.output_filename = output_filename
         self.audio = audio
         self.audio_device = audio_device
+        self.audio_filter = audio_filter
         self.audio_sync = audio_sync
         self.audio_samplerate = audio_samplerate
         self.audio_codec = audio_codec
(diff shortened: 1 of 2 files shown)


davidplowman:
Hi Stephen, thanks very much for this PR. It all looks good to me, I just notice a couple of complaints from flake8 (which I often find to complain over nothing much, but we seem to be using it...). Would you be able to patch those little things up and then I'd be OK to merge. Thanks again!

StephenGrable1:
> Hi Stephen, thanks very much for this PR. It all looks good to me, I just notice a couple of complaints from flake8 (which I often find to complain over nothing much, but we seem to be using it...). Would you be able to patch those little things up and then I'd be OK to merge. Thanks again!

hey @davidplowman sorry for the late follow up here but i finally got around to linting these files - i also added some more context and helpful links to added example, does this still look good to merge in?

picamera2/outputs/ffmpegoutput.py
```diff
@@ -70,7 +71,8 @@ def start(self):
                            '-thread_queue_size', '1024',  # necessary to prevent warnings
                            '-i', self.audio_device]
             audio_codec = ['-b:a', str(self.audio_bitrate),
-                           '-c:a', self.audio_codec]
+                           '-c:a', self.audio_codec,
+                           -'af', self.audio_filter]
```
davidplowman, line 75: Before I merge this, just one thing I wanted to understand. We're now setting the audio filter to `pan=stereo` by default. Is this a change in behaviour, or was it doing the same thing before? (If it is a change, do you know what you would put to get the previous behaviour?) Thanks!

StephenGrable1: Good call, I intended this to be a sane default, but looking into it a bit more and if the input or soundcard is configured for mono, this would indeed change it to stereo (the `-af` default will reflect the input signal/channel setup if nothing is provided to that argument). Its quite rare that a device nowadays would be mono, but im sure there are use-cases for that. 

So I went ahead and removed the default parameter to `pan=stereo` so it will have the same behavior as before, while allowing it to be customized if an arg is passed.

davidplowman: I think we've still got the problem where `audio_filter` is a required parameter, and causing a test failure. Could we put it back where it was (i.e. just before `pts`, just in case anyone out there has hard-coded the order of all those audio parameters) and default it to `None`. Then I think the code you've got after that looks good.

StephenGrable1: done 💯

davidplowman:
Thank you very much!!

PR #1087, opened 2024-08-08
ScalerCrops handling

naushir:


picamera2/picamera2.py
```diff
@@ -1116,6 +1116,15 @@ def configure_(self, camera_config="preview") -> None:
         self.controls = Controls(self, controls=self.camera_config['controls'])
         self.configure_count += 1
 
+        if "ScalerCrops" in self.camera_controls:
+            par_crop = self.camera_controls["ScalerCrops"]
```
davidplowman, line 1120: Ah, I'm guessing `par_crop` means `preserve_ar_crop`? Actually, for my curiosity, how does this work? Presumably libcamera is calculating the AR-preserving crop for each of the cameras - except I thought asking for camera controls was supposed to give us min/max/default triple. So I think I'm a bit confused!

tests/crop_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+import cv2
+
+from picamera2 import Picamera2
+
+picam2 = Picamera2()
+
+for m, l in [(False, False), (False, True), (True, False), (True, True)]:
+    cfg = picam2.create_video_configuration(main={"size": (1920, 1080), "format": 'XRGB8888', "preserve_ar": m},
+                                            lores={"size": (640, 640), "format": 'XRGB8888', "preserve_ar": l},
```
davidplowman, line 13: Just wondering what this does on a Pi 4 where you can't have RGB on the lores?

picamera2/picamera2.py
naushir: > Presumably libcamera is calculating the AR-preserving crop for each of the cameras - except I thought asking for camera controls was supposed to give us min/max/default triple. So I think I'm a bit confused!

This is a bit of a hack where the controlinfo for rpi::ScalerCrops actually gives us the default (i.e. preserved ar) crop for output 0 and output 1 in the min/max fields.  Unfortunately there is no other mechanism that can be used to get this out for span controls.

tests/crop_test.py
naushir: Good question - I assumed the validate() would change this to YUV and everything would be good.  But I've not explicitly tried it...

Actually, there's no reason lores needs to be RGB, perhaps I should just change it to YUV.

picamera2/picamera2.py
davidplowman: I see, yes, that's a bit sneaky! Maybe worth a comment as it was a bit head-scratchy?

PR #1092, opened 2024-08-09
Example: capture_video_timer

mbauer575:
This PR adds a new example script capture_video_timer.py that demonstrates how to capture video for a specific duration starting at a specific time using the picamera2 library. The script includes logging and error handling to ensure smooth operation. 

Usage: 
1. Modify the start_time and duration variables in the script to set the desired start time and recording duration. 
2. Run the script using ```python capture_video_timer.py```

### DIFF
--- a/examples/caputre_video_timer.py
+++ b/examples/caputre_video_timer.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+# Example to capture video for a specific duration starting at a specific time
+# The start time and duration can be modified by changing "start_time" and "duration" variables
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, and thanks for posting this!

Looks good, I guess the main question I had was about the use of threads. Obviously threads can be essential, particularly in more complex applications, though in this case everything could have been run just in Python's main thread. Do you think the threads are adding something in this case? Also (sorry for the nit pick!), I wonder whether the bits of code that start threads should be "joining" them at the end, maybe this would be a bit tidier?

PR #1093, opened 2024-08-12
devices: imx708: Add a new IMX708 helper class

naushir:
This class provides a convenient helper function set enable/disable sensor HDR mode without relying on external tools. Usage example:

```python
from picamera2.devices.imx708 import IMX708

camera_num = 0
c = IMX708(camera_num)
c.set_hdr_mode(True)
picam2 = Picamera2(camera_num)
```

Note that after calling IMX708.set_hdr_mode(), you must re-initialise the Picamera2 instance.

A change to the CameraManager class is also needed to allow it to be reset as it caches sensor modes internally in libcamera.


picamera2/devices/imx708/imx708.py
```diff
@@ -0,0 +1,46 @@
+import fcntl
+import os
+
+from picamera2 import Picamera2
+from v4l2 import *
```
davidplowman, line 5: We should probably just import the bits we need. I'm sure flake8 will complain!

picamera2/devices/imx708/imx708.py
```diff
@@ -0,0 +1,46 @@
...
+HDR_CTRL_ID = 0x009a0915
+
+
+class IMX708:
+    def __init__(self, camera_num: int):
+        self.device_fd = None
+
+        camera_id = next((c['Id'] for c in Picamera2.global_camera_info() if c['Num'] == camera_num), None)
+        if camera_id is None:
+            raise RuntimeError(f'IMX708: Requested camera number {camera_num} cound not be found')
```
davidplowman, line 16: Should we be checking that we do indeed have an imx708? Or selecting it automatically if there's just one of them? Not sure...

picamera2/devices/imx708/imx708.py
```diff
@@ -0,0 +1,46 @@
...
+                if os.path.islink(id_dir) and camera_id in os.readlink(id_dir):
+                    self.device_fd = open(f'/dev/v4l-subdev{i}', 'rb+', buffering=0)
+                    break
+
+        if self.device_fd is None:
+            raise RuntimeError('IMX708: Requested camera device node not found')
+
+    def __del__(self):
+        if self.device_fd:
+            self.device_fd.close()
```
davidplowman, line 32: Does it matter if the device is only closed by the `__del__` method? I might wonder about having a `close()` method that users should be calling. We could also add context manager support so that folks can use `with` and rely on the closing happening automatically.

setup.py
```diff
@@ -30,7 +30,8 @@
         "Programming Language :: Python :: 3.9",
         "Topic :: Multimedia :: Graphics :: Capture :: Digital Camera",
     ],
-    packages=['picamera2', 'picamera2.encoders', 'picamera2.outputs', 'picamera2.previews', 'picamera2.allocators'],
+    packages=['picamera2', 'picamera2.devices.imx708', 'picamera2.encoders', 'picamera2.outputs', 'picamera2.previews',
```
davidplowman, line 33: Do we need this? I thought just `picamera2.devices` would be enough provided you have set up the `devices/__init__.py` (which I think you have).

davidplowman:
Thanks, Naush. Just a couple of general observations... firstly, we should fix the flake8 complaints, and secondly, could we maybe add some kind of a test for this? We could add it to the tests directory (not the examples), and run it automatically, detecting whether we actually have an imx708 attached. It would at least do something to catch any accidental bit-rot!

picamera2/devices/imx708/imx708.py
naushir: > Should we be checking that we do indeed have an imx708?

This is done when traversing the v4l2 subdev nodes below. 

> Or selecting it automatically if there's just one of them?

I'll add this so that camera_num can be optional and it selects the first imx708 sensor found.

PR #1101, opened 2024-08-28
Initial support for Hailo devices

davidplowman:
Add some examples using hefs from /usr/share/hailo-models.

Improve handling of networks with multiple outputs so that they can be used with or without batching. The pose network post-processing is updated slightly to cope with these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ docs/_build/
 .idea
 /.spyproject
 .spyproject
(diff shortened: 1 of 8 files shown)


davidplowman:
@naushir I think I'm in principle ready to merge this if you're happy too...

naushir:
Looks good to me!

PR #1157, opened 2024-11-07
support imx500 to run more output tensors

Andrewymd:
Updated the magic number to 16, as MAX_NUM_TENSORS=8 is too limited for multi-task networks.


davidplowman:
Thanks, I think we actually did this ourselves too. Release is now imminent!

PR #1176, opened 2024-12-12
Correctly annotate PIL Image(s)

pantheraleo-7:
Currently the return types are annotated as `PIL.Image` (the module) and not the intended (?) `PIL.Image.Image` (the class). This PR rectifies that.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1666,7 +1666,7 @@ class Picamera2:
                      partial(capture_arrays_and_switch_back_, self, preview_config, names)]
         return self.dispatch_functions(functions, wait, signal_function, immediate=True)
 
-    def capture_image_(self, name: str) -> Image:
+    def capture_image_(self, name: str) -> Image.Image:
         """Capture image
 
         :param name: Stream name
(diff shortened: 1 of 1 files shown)


pantheraleo-7:
Alternatively we could annotate the return types as `Image.Image` but, currently, there's no need as we only use the class and don't require the module anywhere. But for consistency we could..

pantheraleo-7:
On second thought I would prefer changing the return types rather than the import. Committing the changes.. if the previous simpler approach is desired, we can revert.

pantheraleo-7:
@davidplowman is this PR missing something?

pantheraleo-7:
Yes, it is causing type checker errors.

`PIL` is a library (analogous to a folder)
`PIL.Image` is a module (an `Image.py` file in the `PIL` folder)
`PIL.Image.Image` is a class (this `Image` class represents all the PIL images, it is in the `Image.py` file of the `PIL` folder)

Currently when we return PIL images, we annotate it as `PIL.Image`, so the type checker assumes we are returning the module itself instead of the PIL Image. This causes all kinds of type errors.

pantheraleo-7:
The type check even warns us that a module cannot be annotated as a return type

image

pantheraleo-7:
This can be solved in two ways:
- we directly import the `Image` class using `from PIL.Image import Image`
- we import the module `from PIL import Image` (the convention) and annotate everything with the class `Image.Image`

davidplowman:
OK, thanks for the explanation!

PR #1177, opened 2024-12-12
Suppress very common `configure()` type checker error

pantheraleo-7:
Since the default value was a `str`, passing a `dict` gave type errors. Since the codebase has very minimal static typing [of only return types], this approach is a clean and pythonic way of suppressing the [false-positive] errors without introducing strict static typing of the parameters.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -1130,9 +1130,9 @@ class Picamera2:
                 scaler_crops.append(par_crop[1] if camera_config["lores"]["preserve_ar"] else scaler_crops[0])
             self.set_controls({"ScalerCrops": scaler_crops})
 
-    def configure(self, camera_config="preview") -> None:
-        """Configure the camera system with the given configuration."""
-        self.configure_(camera_config)
+    def configure(self, camera_config=None) -> None:
+        """Configure the camera system with the given configuration. Defaults to the 'preview' configuration."""
+        self.configure_("preview" if camera_config is None else camera_config)
 
     def camera_configuration(self) -> dict:
         """Return the camera configuration."""


davidplowman:
Hi, thanks for the PR. I don't know why there was a test failure when running the change, but I don't think it's related. I may just need to increase a time out or something. But could I ask you to make a couple of easy changes and I think we'll be good to merge this?

1. Could you change the destination branch to raspberrypi/next? We always push new stuff onto the next branch first.
2. Could you squash to two commits together into one? The changes are sufficiently small that I think it will be easier for folks reviewing the history if there's just one. Please could you sign your commit as well (check [here](https://github.com/raspberrypi/picamera2?tab=readme-ov-file#contributing)).

Thanks very much!

pantheraleo-7:
I am sorry that I didn't read the contributing guide.

davidplowman:
> I am sorry that I didn't read the contributing guide.

No need to apologise, always happy to receive contributions! :smiley: 

Tests are all good, so just one thing to check. If wonder if we should add in
```
        if camera_config is None:
            camera_config = "preview"
```
just to 100% guarantee the same behaviour if someone passes in no config at all. Other than that, I'm happy to merge this. Thanks!

pantheraleo-7:
This is what I initially did but if you see [L1047-L1048](https://github.com/raspberrypi/picamera2/blob/f1f9dd249c9a76957aab68a33f0b8245a8466c2f/picamera2/picamera2.py#L1047-L1048), there's already a check for `camera_config=None` which then get set to the preview configuration. So doing this is actually redundant.

davidplowman:
Yes, thanks, I actually noticed that too, but a (very subtle) different thing will happen. In one case you'll get [this](https://github.com/raspberrypi/picamera2/blob/f1f9dd249c9a76957aab68a33f0b8245a8466c2f/picamera2/picamera2.py#L1031), and in the other, the line you pointed out.

To be honest, the whole configuration thing is a bit too twisty these days, but OTOH I don't particularly have the enthusiasm to change  stuff, so it is what it is. If we add that line, then it really is for sure 100% equivalent to what was there before!!

pantheraleo-7:
Is this style okay or do you want a separate if clause?

pantheraleo-7:
Also what's the difference between these two? `self.preview_configuration` is same as `self.create_preview_configuration()` as per [L273](https://github.com/raspberrypi/picamera2/blob/f1f9dd249c9a76957aab68a33f0b8245a8466c2f/picamera2/picamera2.py#L273), does this have to do with [L274](https://github.com/raspberrypi/picamera2/blob/f1f9dd249c9a76957aab68a33f0b8245a8466c2f/picamera2/picamera2.py#L274)?

pantheraleo-7:
I mixed up the `self.*_configuration`s in the `__init__` and the `@property` ones.

Here's the summary:
- "preview" will get you a `CameraConfiguration` object
- None will get you a `dict` object
- The `self.*_configuration`s in the `__init__` ([L273-278](https://github.com/raspberrypi/picamera2/blob/f2d3f162632b9a8e25a5d0a1a701949cadf4fde3/picamera2/picamera2.py#L273-L278)) are redundant because the `@property`s are overwriting them
- They are not only redundant but as they are a `dict` object and _not_ a `CameraConfiguration` object, they are raising an `AttributeError` every time (caught by the generic `except` clause at [L279](https://github.com/raspberrypi/picamera2/blob/f2d3f162632b9a8e25a5d0a1a701949cadf4fde3/picamera2/picamera2.py#L279)) since a `dict` object doesn't have `enable_raw` method

pantheraleo-7:
(I don't have a pi right now, I'll test when I have it and put in a PR if required)

PR #1179, opened 2024-12-17
Resolve various type errors and do some pythonic refactoring of the codebase

pantheraleo-7:
Partly inspired by the edit of [#1177 (Comment)](https://github.com/raspberrypi/picamera2/pull/1177#issuecomment-2548432215)

Type errors resolved in:

1. `close`
- Sets the `camera_ctrl_info` to an empty `dict`, hence keeping the same type. This approach could also be seen in `_reset_flags` [L297](https://github.com/raspberrypi/picamera2/blob/db79106878c85458f769be3d77c251e4549bbe19/picamera2/picamera2.py#L297)
- Sets the `*_configuration` properties to empty `CameraConfiguration` objects, hence keeping the same type.

2. `__init__`
- Initialise the `*_configuration_` attributes instead of setting the `*_configuration` properties which uses the `*_configuration_` attributes under the hood without properly initialising them.

Refactors done in:
1. `_initialize_camera`
- It only ever returned `True` or raised `RuntimeError`, so using a `bool` return type doesn't make sense and is non-pythonic. The downstream usage by `_open_camera` expects a return value of `False` in case of failure, but `False` is never returned.
2. `_open_camera`
- Now it _actually_ catches the error.
- Earlier when `_initialize_camera` returned `True`, the if block was skipped. In the failure case of `_initialize_camera`, the error was propagated, the if block never got the chance to execute.

### DIFF
--- a/picamera2/picamera2.py
+++ b/picamera2/picamera2.py
@@ -270,12 +270,12 @@ class Picamera2:
             _log.debug(f"{self.camera_manager}")
             # We deliberately make raw streams with no size so that it will be filled in
             # later once the main stream size has been set.
-            self.preview_configuration = self.create_preview_configuration()
-            self.preview_configuration.enable_raw()  # causes the size to be reset to None
-            self.still_configuration = self.create_still_configuration()
-            self.still_configuration.enable_raw()  # ditto
-            self.video_configuration = self.create_video_configuration()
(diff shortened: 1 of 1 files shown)


pantheraleo-7:
I think I have understood what's happening.

Issue #1192, opened 2025-01-20
[SUGGESTION] Add typings and support for local development


KarimAziev:
**Is your feature request related to a problem? Please describe.**
Picamera2 currently lacks type annotations for many of its methods, making it difficult for developers to understand method signatures, expected inputs, and return types.

**Additional context**
After buying the Camera Module 3 as a replacement for Camera Module 2, I discovered that Camera Module 3 is **not backward compatible** with the V4L2/MMAL stack. This was **never mentioned in the official documentation**.

So now developers have no choice but to rewrite their code to use **picamera2**. However, the experience of using picamera2 is not the best due to the lack of type annotations and inline documentation, at least for the public methods.

Take, for example, the `capture_array` method, a highly common method:
```python
def capture_array(self, name="main", wait=None, signal_function=None):
        """Make a 2d image from the next frame in the named stream."""
        return self.dispatch_functions([partial(self.capture_array_, name)], wait, signal_function)
```

Based on its name (and the minimal docstring), it’s supposed to return the captured frame as a 2D array. Yet, the return type is `None` in your dispatch code:
```python
def dispatch_functions(self, functions, wait, signal_function=None, immediate=False) -> None:
```

And this is not an isolated example. Another core method, `create_video_configuration`, suffers from similar problems:

```python
def create_video_configuration(self,
                               main={},
                               lores=None,
                               raw={},
                               transform=libcamera.Transform(),
                               colour_space=None,
                               buffer_count=6,
                               controls={},
                               display="main",
                               encode="main",
                               queue=True,
                               sensor={},
                               use_case="video") -> dict:
```

What are the valid keys in the `main`, `lores`, or `raw` dictionaries? What is `"lores"` even supposed to mean?

What are the possible `"format"` values (`XBGR8888`, `YUV420`, etc.)? Can developers query these programmatically?

Yes, I saw the PDF documentation with descriptions but jumping every time to the PDF documentation is not practical. Documentation should be directly accessible in the code.

I am not exaggerating, because the importance of proper typings grows due to the impossibility of local development without Raspberry Pi hardware, unlike the legacy MMAL/Camera Stack.

There are no mocks, no fallback mechanisms, nothing to let developers stub or test their code.

How can developers expect to build reliable applications when they’re forced to test everything live on Raspberry Pi hardware and without at least proper typings and inline documentation?

**Describe the solution you'd like**

- Include proper type annotations to all public methods for arguments and return types.
- Improving inline documentation.
[...]

PR #1195, opened 2025-01-22
ci: Add Pyright static type checking workflow

KarimAziev:
This pull request introduces Pyright, a static type-checking tool for Python, into the GitHub Actions CI pipeline. 

As part of addressing issue [#1192](https://github.com/RaspberryPi/picamera2/issues/1192), this addition encourages gradual improvements in type annotations while providing developers with immediate feedback on type-related issues.

At this stage, the pipeline warns about type issues and doesn’t enforce strict checking. This is intentional to allow gradual type annotation improvements over time.

### DIFF
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -1,0 +1,25 @@
+name: Pyright static type checking
+
+on:
+  pull_request:
+    branches: ['*']
+  workflow_dispatch:
+
+jobs:
+  type-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Node.js (needed for Pyright)
+        uses: actions/setup-node@v4
+        with:
+          node-version: current
+
+      - name: Install Pyright
+        run: npm install -g pyright
+
+      - name: Run Pyright static type checker
+        run: pyright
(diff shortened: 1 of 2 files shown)


davidplowman:
Thanks, that's really great! One further thought... do you know if it would be possible to install this check as part of our pre-commit hook? Not essential, I guess, but it might be nice to be told about errors before creating a pull request.

KarimAziev:
Yes, you're correct that it is possible to integrate Pyright into the pre-commit hooks via the `.pre-commit-config.yaml` file. However, it might be too early to consider adding it to the pre-commit workflow at this stage.

When I initially added Pyright to the CI for this repository, it detected around 400 type-related issues. To avoid blocking progress, these issues were downgraded to warnings for now, providing visibility without enforcing strict type-checking.

Adding Pyright as a strict pre-commit check at this point might create unnecessary friction for contributors, as many existing warnings could interfere with commit workflows. An alternative approach could be to encourage developers to configure their IDEs to use Pyright, allowing them to see and fix warnings incrementally while working on the codebase.

A potential strategy could involve enabling stricter checks incrementally. For example, starting with `reportReturnType` as an error (which would raise ~20 manageable errors) could help enforce type safety gradually without overwhelming contributors.

Once the majority of issues are addressed and warnings are cleaned up, it might make sense to revisit the idea of adding Pyright to the pre-commit hooks as a way to enforce stricter type-checking for new changes. For now, though, an incremental improvement process seems more practical.

davidplowman:
Yes, those are good points. Maybe I'll just start using it locally, figure out how to start enabling just a few things, and see how it goes for a bit. Thanks!

PR #1198, opened 2025-01-24
Add PySide6 Compatible Picamera2 Widgets

LODean:
Widgets "QGlPicamera2" and "QPicamera2" should be imported from from "picamera2.previews.qt_pyside6", rather than "picamera2.previews.qt", for use with PySide6.

Widgets can be used within a PySide6 application the same as they would with PyQt5.

Example of using QGLPicamera2 widget with PySide6 (modified from example in Picamera2 manual):
```
from PySide6.QtWidgets import QApplication
from picamera2.previews.qt_pyside6 import QGlPicamera2

picam2 = Picamera2()
picam2.configure(picam2.create_preview_configuration())

app = QApplication()
qpicamera2 = QGlPicamera2(picam2, width=800, height=600, keep_ar=True)
qpicamera2.setWindowTitle("PySide6 Qt GL Picamera2 App")

picam2.start()
qpicamera2.show()
app.exec()
```

A QPicamera2 widget (no GPU integration) can be used identically by simply replacing "QGlPicamera2" with "QPicamera2".

### DIFF
--- a/picamera2/previews/q_gl_picamera2.py
+++ b/picamera2/previews/q_gl_picamera2.py
@@ -79,8 +79,8 @@ class QGlPicamera2(QWidget):
         super().__init__(parent=parent)
         self.resize(width, height)
 
-        self.setAttribute(Qt.WA_PaintOnScreen)
-        self.setAttribute(Qt.WA_NativeWindow)
+        self.setAttribute(Qt.WidgetAttribute.WA_PaintOnScreen)
+        self.setAttribute(Qt.WidgetAttribute.WA_NativeWindow)
 
         self.bg_colour = [colour / 255.0 for colour in bg_colour] + [1.0]
         self.keep_ar = keep_ar
@@ -108,7 +108,7 @@ class QGlPicamera2(QWidget):
         self.preview_window = preview_window
 
         self.camera_notifier = QSocketNotifier(self.picamera2.notifyme_r,
-                                               QSocketNotifier.Read, self)
+                                               QSocketNotifier.Type.Read, self)
         self.camera_notifier.activated.connect(self.handle_requests)
         # Must always run cleanup when this widget goes away.
         self.destroyed.connect(lambda: self.cleanup())
--- a/picamera2/previews/q_picamera2.py
+++ b/picamera2/previews/q_picamera2.py
@@ -36,14 +36,14 @@ class QPicamera2(QGraphicsView):
         self.setScene(self.scene)
         self.setBackgroundBrush(QBrush(QColor(*bg_colour)))
         self.resize(width, height)
-        self.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
-        self.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self.setHorizontalScrollBarPolicy(Qt.ScrollBarPolicy.ScrollBarAlwaysOff)
+        self.setVerticalScrollBarPolicy(Qt.ScrollBarPolicy.ScrollBarAlwaysOff)
         self.enabled = True
         self.title_function = None
 
         self.update_overlay_signal.connect(self.update_overlay)
         self.camera_notifier = QSocketNotifier(self.picamera2.notifyme_r,
-                                               QSocketNotifier.Read, self)
+                                               QSocketNotifier.Type.Read, self)
         self.camera_notifier.activated.connect(self.handle_requests)
         # Must always run cleanup when this widget goes away.
         self.destroyed.connect(lambda: self.cleanup())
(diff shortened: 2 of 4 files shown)


davidplowman:
Hi, and thanks very much for doing this, it looks really interesting! I'd like to merge this but have a couple of questions.

* Do you think there's a way to package this that could avoid the duplication of entire files? Maybe one could have just one implementation file, but change what it imports?
* (Minor) Before we finally merge, it would be nice to have an example in the examples folder, and to added it to the CI tests.

Thanks again!

Akshar062:
If this is going to get merge it would be nice to also add support for pyside2 (it is not too much work so i think it would be best ) and can this get merge ?

LODean:
Hi,

I have been considering what the best way forward for this might be, though there are a few things to take into account. We do need to reduce duplicated code, and allow for addition of different Qt GUI Python implementations, but it also needs to be kept compatible with current implementations (e.g. a QPicamera2 call using PyQt5 by default).

I was also thinking it is best to not rely on there being only one Python Qt implementation on the environment, and therefore having a order of precedence on the import of Qt implementations isn't ideal, it is best if it can be selected.

I'm working on a solution for these issues, would you think it necessary to allow for different Qt implementations to be used within the same Python program?

davidplowman:
Hi, and thanks for the reply. I can't say that I see any need for people to run different versions of PtQt (or whatever) in the same process, so my feeling is that we don't need to worry about that. That sounds fair to me, but I'm open to persuasion if anyone disagrees! :)

So anyway, I think that means that some kind of scheme where maybe it tries multiple alternatives - but ideally lets you choose if you want - sounds promising?

LODean:
I have found a solution for these issues and am in the process of properly implementing it.

davidplowman:
Great, keep us posted, and thank you very much!

davidplowman:
Hi, so there were a couple of test failures there. I don't think the flake8/lint ones are terribly major and should be easy to clear up. You should be able to get flake8 running as a pre-commit check with something like
```
pip install pre-commit
pre-commit install
```
The static type checking is a work-in-progress. Another contributor is very kindly helping to improve the typing hints, but there's a lot to do so we're a bit in the middle of things. But I think it's probably only the two errors at the end that matter at the moment. If you have `pyright` installed, I think you might be able to get just the errors if you run `python -m pyright --level error`. Eventually we'll get that into the pre-commit checks too. But sorry that this is all a bit painful at the moment.

LODean:
I believe I have now corrected those errors (which were both type checking errors too), and also corrected the linting.

This should now be compatible with PyQt5, PyQt6, PySide2, and PySide6, though only in widget format which will need to be added to your own application (the default preview window will still attempt to use PyQt5).

Classes can be imported from **picamera2.previews.qt** as before for specific versions use classes as follows:

| Qt Implementation     | PyQt5               | PyQt6                | PySide2                  | PySide6                    |
|-------------------------|-----------------|-------------------|----------------------|-----------------------|
| Standard Qt Widget    | QPicamera2    | Q6Picamera2     | QSide2Picamera2    | QSide6Picamera2    |
| Hardware Accelerated | QGlPicamera2 | QGl6Picamera2 | QGlSide2Picamera2 | QGlSide6Picamera2 |

All classes can be interacted with in identical ways, except for the different Qt implementation in their application.

These classes are lazy loaded in picamera2.previews.qt so may not show up in syntax highlighting as they are dynamically accessed whereas syntax highlighting often relies on static analysis.

LODean:
Also, I have not yet added examples to the examples folder or added tests for each new class to the the CI tests, I am planning on doing so at some point but can do a new pull request for that once I have done so, if that helps.

davidplowman:
Thanks, that's looking good! So if I'm running a different GUI framework, all I do is replace
```
from picamera2.previews.qt import QGlPicamera2
```
by
```
from picamera2.previews.qt import QGlSide6Picamera2
```
(or whichever). Obviously the rest of the GUI code is not the camera system's concern. 

I guess I should just have a read through to make sure I understand what's going on, and also I should try it out on at least one of the other frameworks, so maybe just give me a day or so to do that.

Final question... if one did want the automatic preview to select the flavour of Qt that is installed, do you have any ideas on that? Maybe I can give that some thought too once I've got a bit more of an understanding. Thanks again!

LODean:
Yes you just replace the import as you stated,

In terms of the automatic preview you'd probably want something similar to what is done for the QPicamera2 class as the QtPreviewBase class is dependent on the imported Qt implementation. Duplication of that class for each implementation would also work but would again result in large amounts of code duplication.

davidplowman:
Sure! Let me try all this first, and hopefully merge tomorrow. And we can leave other stuff on the back burner just for the moment!

Akshar062:
I have tried with pyside2 and 6 it is working wonderfully , but I try many thing but i can't get it work with QML(qt 5) and pyside2 , i know I am asking lot but if you know about qml and can make one example it will be big help ( i have to use it until i make new app in pyside6)

davidplowman:
I've tried out PyQt6 and PySide6 with a little test app based on [this one](https://github.com/raspberrypi/picamera2/blob/main/tests/app_test.py). Both worked perfectly (ignoring a minor sytnax change to the label alignment).

Just as a note for anyone else, to make the PySide6 version work, I had to do
```
sudo apt install libxcb-cursor0
```
before it would run.

(I can probably add those to the CI tests which will ensure we never break this new functionality completely.)

PR #1209, opened 2025-02-13
Copy optimise

davidplowman:
This PR refactors some of the copying out of image data from requests. With various improvements to PyAV and simplejpeg we can now avoid copies in many more cases. It also provides access directly to simplejpeg's new YUV encode feature, meaning we can allocate much smaller buffer for image encoding.


mgineer85:
Hey @davidplowman , 

I would like to raise again the accidental change that always evaluates False:
https://github.com/raspberrypi/picamera2/pull/1209/files#diff-26e92d0368f682e2f447e8e5c597e95d27f67eff88e4345d3d4fff4226eb8fcb

line 1163:  ` if False and "ScalerCrops" in self.camera_controls:`

I think ``False and`` shall be removed.

davidplowman:
Oh dear, is that still in there. I'll fix it.

(It was so annoying that libcamera got completely broken at one point so that this was necessary for anything to work...!)

PR #1257, opened 2025-04-30
Add 180 to mjpeg server with rotation example

codewrite:
Signed-off-by: Jon Nethercott <jon@codewrite.co.uk>

### DIFF
--- a/examples/mjpeg_server_with_rotation.py
+++ b/examples/mjpeg_server_with_rotation.py
@@ -14,14 +14,15 @@ from picamera2 import Picamera2
 from picamera2.encoders import MJPEGEncoder
 from picamera2.outputs import FileOutput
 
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, and thanks for the PR. I don't know why it doesn't already handle 180 degree rotation, I guess the thought is that you can do that by setting up the camera with a 180 degree rotation transform, and you get a "truly" rotated image, rather than one that's only rotated by metadata. But you're right - why not complete the set here as well?

davidplowman:
Well, I'm a bit baffled what's happened to the flake8 linting there, it was working yesterday! Some unexplained dependency problem or bitrot, who knows? *Sigh.* But I can't see that it's related to this PR in any way,

PR #1262, opened 2025-05-06
Fix setting of recording to False in stop() call of CircularOutput2

mbirghan:
Right now stoping CircularOutput2 does not correctly set recording to False.

All other instances use `self.recording` while here `self._recording` was used.


davidplowman:
Thanks very much for this!

PR #1279, opened 2025-06-13
imx500: Add support for Device ID and encrypted rpks

naushir:
Allow the imx500 library to readback the device id from the kernel using a V4L2 control.

Additionally parse the encryption flags (if present) in the rpk binary.


naushir:
@roliver-rpi FYI

roliver-rpi (approved):

davidplowman:
It all seems perfectly fine to me... I'll have to take your word for it that it actually does the right thing!!

I'll merge now - do we want to be doing a release soon?

naushir:
Thanks!
Any possibility of making a package release with this change in the coming weeks?

Issue #1300, opened 2025-07-28
[SUGGESTION] HailoDevice context manager for HailoModel


gh9869827:
## Is your feature request related to a problem? Please describe.

Not a blocking issue, but the current `HailoModel` implementation in `picamera2/devices/hailo/hailo.py` uses static class-level variables (`TARGET`, `TARGET_REF_COUNT`) to manage the shared Hailo virtual device. This approach does support using multiple models on a single device, but:

- It hides lifecycle management behind global state.
- It prevents users from explicitly controlling when the virtual device is created or released.

## Describe the solution you'd like

Refactor the code to:

- Introduce a `HailoDevice` class to explicitly manage the virtual device, with support for context-managed lifetime (`with HailoDevice() as device:`).
- Update `HailoModel` to take a `device: HailoDevice` as a parameter instead of relying on global variables.
- Optionally, use a `HailoModelConfig` NamedTuple to encapsulate model parameters (`hef_path`, `output_type`, `batch_size`).

This preserves all current functionality, including support for multiple models on the same device, while introducing a mechanism to explicitely control the lifecycle of the device using a `with` statement, independantly of the `HailoModel`s.

## Additional context

I've attached a proposed implementation for the mechanism I describe above. It includes header comments explaining the changes and preserves attribution to the original Picamera2 source.

I'd be happy to open a pull request if there's interest in this direction.

[hailo_py.txt](https://github.com/user-attachments/files/21478258/hailo_py.txt)

davidplowman:
Hi, and thanks for the suggestion. In principle I'm keen to accept contributions that will make things work better, just a couple of questions.

The existing code has `__enter__`/`__exit__` context manager methods, so are we replacing these? My impression with them was that they don't work correctly with multiple models - presumably this will fix that? We can "deprecate" those old methods with a warning, and indicate to users what the new replacement code would be.

We should obviously also add an example of this updated API to the repo so that it's clear how to use it. There's also a multiple model demo [here](https://github.com/raspberrypi/face_recogniser), perhaps we could update that to use this new context manager and demonstrate that it works correctly now?

I feel slightly that it might be worth making the changes that give us functional fixes first, and leave the cosmetic config changes for another time. I'm just always really nervous about changing more more stuff under our user's feet in one go than is absolutely necessary. Does that sound OK?

[...]

PR #1303, opened 2025-07-30
Fix incorrect Hailo reference counting and virtual device reuse

gh9869827:
## Fix incorrect Hailo reference counting and virtual device reuse

This PR fixes incorrect reference tracking of the shared virtual device in the `Hailo` class when used with multiple model instances.

### Problem

- `TARGET_REF_COUNT` was only incremented during the **first** virtual device allocation.  
  As a result, loading multiple models caused the `VDevice` to be **deallocated prematurely**, since each model called `.close()` and decremented the counter.

- `TARGET` was **not reset to `None`** after releasing the virtual device, which caused a released `VDevice` to be reused on the next model creation, resulting in runtime exceptions.

### Solution

- Moved the `TARGET_REF_COUNT += 1` line **outside** the `if TARGET is None` block to ensure it increments for *every* model.
- Reset `TARGET = None` after releasing the virtual device when the count reaches zero, allowing future instances to allocate a new one properly.

#### Test program illustrating premature deallocation

```python
from hailo import Hailo

m1 = Hailo("yolov8l.hef")
m2 = Hailo("yolov8l.hef")

with (m1, m2):
    pass  # before fix: Segfault
```

#### Test program illustrating reuse of released device

```python
from hailo import Hailo

with Hailo("yolov8l.hef") as m1:
    pass

with Hailo("yolov8l.hef") as m1:
    pass  # before fix: AttributeError: 'NoneType' object has no attribute 'create_infer_model_from_file'
```

### Summary of changes

- Adjusted reference count logic in `__init__` and `close`
- Added inline comments for clarity
- No change in external behavior

### Notes

This PR addresses the problems discussed in [Issue #1300](https://github.com/raspberrypi/picamera2/issues/1300).

### DIFF
--- a/picamera2/devices/hailo/hailo.py
+++ b/picamera2/devices/hailo/hailo.py
@@ -25,7 +25,8 @@ class Hailo:
         self.hef = HEF(hef_path)
         if Hailo.TARGET is None:
             Hailo.TARGET = VDevice(params)
-            Hailo.TARGET_REF_COUNT += 1
+        # increment `TARGET_REF_COUNT` every time a model is loaded
+        Hailo.TARGET_REF_COUNT += 1
         self.target = Hailo.TARGET
         self.infer_model = self.target.create_infer_model(hef_path)
         self.infer_model.set_batch_size(1 if batch_size is None else batch_size)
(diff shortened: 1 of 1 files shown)


davidplowman:
Thanks very much for this. I'm good to merge this - I've tested it and confirmed that it does indeed fix this problem.  The only very minor thing is that a "Signed-off-by:" would be great if you're OK to add that to the commit message.  Thanks again!

gh9869827:
Hi, Absolutely, I've added the `Signed-off-by:` line and force-pushed the updated commit. Let me know if there's anything else you'd like me to adjust!

davidplowman:
Thank you! :)

PR #1326, opened 2025-09-29
Add Remote Requests

PineappleBeech:
Let requests be handled in child processes

### DIFF
--- a/examples/remote_processing_example.py
+++ b/examples/remote_processing_example.py
@@ -1,0 +1,55 @@
+#!/usr/bin/python3
+
+# Shows how to use the remote module to process images in a separate process.
+# This is doing a simple sum of the pixels in the image in a separate process.
+# The results are then gathered by the main process and printed.
+
+import multiprocessing as mp
+import os
+import queue
+import threading
+import time
+
+import numpy as np
+
+import picamera2
+from picamera2 import Pool, RemoteMappedArray
+
+IMAGE_COUNT = 100
+
+
+def run(request):
+    print(f"Recieved request in process {os.getpid()}")
+    with RemoteMappedArray(request, "main") as m:
+        s = np.sum(m.array)
+        time.sleep(0.1)  # long calculation
+        print(f"Sum of array: {s}")
+    print(f"Processed request in process {os.getpid()}")
+    return s
+
+
+def gather_results(futures):
+    for i in range(IMAGE_COUNT):
+        future = futures.get()
+        print(f"Recieved result {i + 1}: {future.result()}")
+
+
+if __name__ == "__main__":
+    mp.set_start_method("spawn")
+    picam2 = picamera2.Picamera2()
+    config = picam2.create_preview_configuration(buffer_count=16)
+
+    with Pool(run, 16, picam2) as pool:
+        # Start the camera after creating the pool so the camera doesn't timeout
+        picam2.start(config)
+
+        futures = queue.Queue()
+        results_thread = threading.Thread(target=gather_results, args=(futures,))
+        results_thread.start()
+
+        for _ in range(IMAGE_COUNT):
+            with picam2.captured_request() as request:
+                future = pool.send(request)
+                futures.put(future)
+
+        results_thread.join()
--- a/picamera2/remote.py
+++ b/picamera2/remote.py
@@ -1,0 +1,429 @@
+import mmap
+import multiprocessing as mp
+import os
+import queue
+from collections import deque
+from concurrent.futures import Future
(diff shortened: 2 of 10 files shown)


picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+import numpy as np
+from libcamera import ColorSpace, Transform
+from PIL import Image
+
+import picamera2
+
+
+class Process:
+    """
+    Creates a new process which recieves requests from the parent process.
```
davidplowman, line 20: receives (sorry!)

picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+                timeout_counter += 1
+                if timeout_counter >= self._timeout:
+                    _, future = self._requests_sent.popleft()
+                    future.set_exception(TimeoutError("No result received for timeout period"))
+                continue
+
+            timeout_counter = 0
+
+            if not bool(self._requests_sent):
+                break
```
davidplowman, line 74: Just wanted to check... should this be a `continue`?

picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+        self._thread.start()
+
+        self._process = _RemoteProcess(self._send_queue, self._return_queue, picam2, run, init)
+
+    def _return_thread(self):
+        """Thread that passes the return values to the Future objects."""
+        timeout_counter = 0
+        while True:
+            try:
+                result = self._return_queue.get(timeout=1 if self._timeout is not None else None)
```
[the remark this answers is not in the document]
PineappleBeech: That causes the process to return result for each value in its queue first so it's not instant but means every future gets a result before it exits.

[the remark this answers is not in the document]
davidplowman: Understood!

picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+        return self._array
+
+
+class RemoteRequest:
+    """
+    A request that is sent to the child process.
+
+    Args:
+        request: The CompletedRequest object
+    """
```
davidplowman, line 223: I have a feeling we might need to think about what happens if someone writes back to the underlying camera buffer. Do we need DMA buf syncs? (i.e. flush the data cache back to memory)

If this is too painful, we could make buffers read-only.

picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+        array = np.array(mem, copy=False, dtype=np.uint8)
+        config = self.config[name]
+        width, height = config["size"]
+        stride = config["stride"]
+        format = config["format"]
+        if format == "YUV420":
+            return array.reshape((height + height // 2, stride))
+        array = array.reshape((height, stride))
+        if format in ("RGB888", "BGR888"):
+            return array[:, :width * 3].reshape((height, width, 3))
```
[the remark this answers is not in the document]
davidplowman: Sounds good. We should probably double-check we have at least one of each of these types (YUV420, 24-bit, 32-bit) in a test somewhere!

picamera2/remote.py
```diff
@@ -0,0 +1,429 @@
...
+                array = np.copy(m.array)
+                stride = array.strides[0]
+            elif not m.array.data.c_contiguous:
+                array = m.array.base
+            img = Image.frombuffer("RGB", (shape[1], shape[0]), array, "raw", mode, stride, 1)
+        return img
+
+    def save(self, filename: str, stream_name: str = "main"):
+        """Save an image to a file."""
+        img = self.make_image(stream_name)
```
davidplowman, line 344: I'm wondering just a little bit whether we should try and save some EXIF stuff for JPEGs? Possibly also respect the quality/compress_level values in the options if that's sufficiently straightforward?

PineappleBeech: I've made it use the Helpers.save() method which will add EXIF data. the quality/compress_level are copied from the Picamera2 object when the process in created so it will now use whatever they were set to then

examples/remote_processing_example.py
```diff
@@ -0,0 +1,54 @@
...
+        future = futures.get()
+        print(f"Recieved result {i + 1}: {future.result()}")
+
+
+if __name__ == "__main__":
+    mp.set_start_method("spawn")
+    picam2 = picamera2.Picamera2()
+    config = picam2.create_preview_configuration(buffer_count=16)
+
+    with Pool(run, 16, picam2) as pool:
```
davidplowman, line 42: Maybe we should put in a comment that we recommend opening the pool before starting the camera, as process creation is expensive? Ditto for the other similar examples.

PR #1330, opened 2025-10-09
imx500: Implement input tensor injection and demo

roliver-rpi:


picamera2/devices/imx500/imx500.py
```diff
@@ -319,6 +326,7 @@ def __init__(self, network_file: str, camera_id: str = ''):
             self.fw_progress_chunk = open(f'/sys/kernel/debug/rp2040-spi:{spi_device_id}/transfer_progress', 'r')
 
         if self.config['network_file'] != '':
+            self.__enable_injection()
```
naushir: This happens unconditionally?

roliver-rpi: Thanks for catching this. This has inadvertently made it through from an earlier stage of development.

PR #1336, opened 2025-11-03
imx500: Support BGR/unsigned input tensors

roliver-rpi:


roliver-rpi:
@davidplowman , do you know if the pi4/5 runners are going this morning?

PR #1372, opened 2026-03-25
tests: imx500: More robust imx500 testing

naushir:
Don't read the ID until we are sure the camera has started streaming.


naushir:
Seems to be more stable for one class of test failures!

PR #1376, opened 2026-04-10
Check camera is be configured in start_encoder()

davidplowman:


naushir:
@davidplowman tests now passing.

PR #1379, opened 2026-04-22
imx500: Allow firmware progress bar to work for non-sudo users

davidplowman:
We add a couple of little scripts to read the fw update status, and allow them to run for everyone with password-less sudo.

Please see imx500_progress_bar.txt for instructions on how to install and use this feature.


davidplowman:
@naushir Are we happy with this?

naushir (approved):

PR #1383, opened 2026-05-05
app_full.py is compatible with monochrome sensors. Fixes #1380

### LINKED ISSUE #1380: [BUG] app_full.py example using monochrome sensor
just a minor remark here..

the app_full.py example works fine with a color sensor, but crashes when using a monochrome sensor with Y10 pixel format.

Simple workaround seems to be checking if there is a 'ColorFilterArray' property in the camera_properties, before setting saturation/awb/ccm controls.

fiepfiep:
Fixes #1380
add a check for monochrome sensors before accessing controls such as AWB, saturation.
@davidplowman

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 
 ### Changed
 
+## 0.3.35 Beta Release 34
+
+### Added
+
+### Changed
+
+* IMX500 progress bar can be made to work for non-sudo users.
+* Formatting check has been changed over to Ruff.
+* Other minor fixes (see commit history).
+
 ## 0.3.34 Beta Release 33
 
 ### Added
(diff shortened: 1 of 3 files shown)


davidplowman:
Thanks for the PR. Actually we've just added a mono sensor into our "fleet" of test devices which is why there are some failures, so we'll get those cleaned up first.

On reflection, I might add a `picam2.is_mono()` function, which might make this a bit neater. I'll also replace `5` by `libcamera.properties.draft.ColorFilterArrangementEnum.MONO` which, although a bit of a mouthful, doesn't have that fragility problem we mentioned elsewhere. Thanks!

davidplowman:
We've gone through all the other test failures for mono sensors and fixed them up, and taken care of this one too. That's all been merged onto our `next` branch now so will be in the next release, along with a `picam2.is_mono` property. I think that takes care of everything so I'll close this one now, but thanks for bringing it all to our attention. With a mono sensor in our test farm too now, hopefully we'll get less of this kind of thing in future!

PR #1386, opened 2026-05-12
Speedup import time

nicolasdespres:
Importing OpenCV on SD card is slow (~15sec).

This patch removes OpenCV import from the main path to speed up import time of imx500.

An alternative solution would be to use another library like Pillow which is faster to import and provide similar drawing performance.

### DIFF
--- a/picamera2/devices/imx500/postprocess.py
+++ b/picamera2/devices/imx500/postprocess.py
@@ -10,7 +10,6 @@ https://github.com/stefanopini/simple-HigherHRNet
 from enum import Enum
 from typing import List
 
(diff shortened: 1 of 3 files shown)


davidplowman:
Hi, thanks for this PR, this seems like a good idea. We're using Ruff now for code style checking and it had just a few little complaints - I think all it needs is to add a blank line after the new imports. Would you be OK to fix that and then I'm happy to merge when all the tests are happy. Thanks!

Edit: Oh, and please do also add a `Signed-off-by: Name <email address>` line to the commit message if you haven't done so already!

PR #1393, opened 2026-06-03
tests: aegc: Fix possible false postive failure

naushir:
If the auto algorithm converges on the same gain/exposure as the last manual value, the test will fail. Try and detect this case and do one more manual control step before switching back to auto mode.

### DIFF
--- a/tests/aegc.py
+++ b/tests/aegc.py
@@ -22,9 +22,20 @@ def test_control_auto(control):
     for _ in range(5):
         check = picam2.capture_metadata()[control]
         if abs(check - current) > current * 0.05:
-            print(f"Control {control} changed from {current} to {check}")
-            return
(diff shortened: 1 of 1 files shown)


tests/aegc.py
```diff
@@ -22,9 +22,20 @@ def test_control_auto(control):
...
+def run_auto_test(control, reset_value):
+    success, current, check = test_control_auto(control)
+    if not success:
+        # Retry in case the auto algorithm converged back to the last fixed value
+        test_control_fixed(control, reset_value)
+        success, current, check = test_control_auto(control)
+    if success:
+        print(f"Control {control} changed from {current} to {check}")
+    else:
+        print(f"ERROR: {control} has not returned to auto - still {check}")
```
davidplowman: Just wondering if maybe it should be `still {current}`, but otherwise LGTM!

